Issue #11, opened 2016-02-09
Support calling GDB over SSH


bbenoist:
Could it be possible to implement support for calling GDB over SSH ?

From what I've understood, it seems quite easy to implement within code-debug. You would simply need to replace the system executable calls and stdio redirections by an appropriate SSH implementation. [simple-ssh](https://www.npmjs.com/package/simple-ssh) seems to do the job but there might be better alternatives.

You can take a look at the [Visual Studio (!= vscode) GDB Debugger Extension](https://blogs.msdn.microsoft.com/vcblog/2015/11/18/announcing-the-vs-gdb-debugger-extension/) for an example implementation of this use-case. Source code of [Microsoft/MIEngine](https://github.com/Microsoft/MIEngine) might also help :wink: 

Here are my personal requirements for this feature:
* Use either authentication with a password or with an SSH private key file.
* Expose the possibility to use a custom SSH port
* The `target` field must point to the target executable not the ssh server

`launch.json` proposal:
```json
"request": "launch",
"target": "./bin/executable",
"cwd": "/home/target/build",
"ssh": {
  "host": "127.0.0.1",
  "port": "2222",
  "user": "jdoe",
  "password": "pw4jdoe"
}
```

WebFreak001:
Saving a password in clear text is terrible though. This should only allow private key files

WebFreak001:
Maybe "target": "jdoe@127.0.0.1:2222:/home/target/build/bin/executable" instead

bbenoist:
Thanks for the feedback :+1: 

> Saving a password in clear text is terrible though.

Maybe for the most common usages but not when using a [Vagrant](https://www.vagrantup.com/) box whose SSH ports are only exposed to the host system and whose SSH private keys are automatically regenerated :wink: 

> This should only allow private key files

I tend to believe that such decisions should be left to the end-user.

> Maybe "target": "jdoe@127.0.0.1:2222:/home/target/build/bin/executable" instead

If you give me the choice between an option requiring a specific string format and an object consisting of clearly separated options, I would choose the first.

I find it worse that my configuration proposal because it would involve extra considerations which are not actually required. Apart from a debatable removal of extra JSON options (you would still have to specify a private key or password however), I see no obvious reason for creating an extra string format whereas such options would end up being loaded separately in the code.

WebFreak001:
Right, and having children in the json also gives better auto completion for vscode. Might use that then instead

bbenoist:
[...]

PR #13, opened 2016-02-10
Fix invalid local windows path when using GDB over SSH

bbenoist:
Because of the global POSIX configuration for path-related processing, the
local absolute paths of files with a breakpoint were not translated
correctly.

This fix forces the use of the default NodeJS path.resolve instead of the
global POSIX override.

This PR should fix the issues mentioned in #11

### DIFF
--- a/src/gdb.ts
+++ b/src/gdb.ts
@@ -5,6 +5,7 @@ import { MINode } from './backend/mi_parse'
 import { expandValue, isExpandable } from './backend/gdb_expansion'
 import { MI2 } from './backend/mi2/mi2'
 import { posix } from "path"
+import * as systemPath from "path"
 
 let resolve = posix.resolve;
 let relative = posix.relative;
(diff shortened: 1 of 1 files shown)


bbenoist:
:+1:

@WebFreak001 Could you please release a 0.3.2 version?

Having the possibility to remote-debug their programs within vscode and from Windows is enough to justify it IMO :wink:

There seems to be numerous people which might be interested in it:
* https://visualstudio.uservoice.com/forums/293070-visual-studio-code/suggestions/7752459-c-code-completion-and-debugging
* http://stackoverflow.com/questions/35008520/configuring-vs-code-with-gdb
* https://github.com/Microsoft/vscode/issues/1627

bbenoist:
@WebFreak001 Thank you for https://github.com/WebFreak001/code-debug/commit/186bb689557d0103a1e52bfef46a96e0adf78bd0 !

HaronK:
Hi,
I found out that the similar code that was fixed by this PR present in other function (setBreakPointsRequest) in the same file (now it's called mibase.ts).
In my case VSCode is running on Windows and remote code is on Linux.
So when I put breakepoint in the code and start debugging extension shows error message box with path to the file consisted of part of path from Windows cwd and rest from ssh.cwd and file name. And of course it cannot find this file.
I suspect that this is caused by code I mentioned above so if someone can check and fix this it would be very nice.
Thanks in advance.

Issue #15, opened 2016-02-10
Add option to execute extra commands before running GDB when using SSH


bbenoist:
For some reasons, I need to execute an extra command (loading an appropriate execution environment on the target machine) before executing GDB over SSH.

Could it be possible to add an `extraCommands` options in the config which would contains the commands to execute on the remote machine (within the same SSH session than GDB) before calling GDB?

WebFreak001:
the ssh package im using does not really support that. Instead you could make a script on the remote machine and make it get called when GDB is called

bbenoist:
> the ssh package im using does not really support that. Instead you could make a script on the remote machine and make it get called when GDB is called

See https://github.com/bbenoist/code-debug/commit/d6ee383e75ead10de539db063cba5063014bb4af []() for an example implementation responding to my needs.
I simply added an additional `ssh.bootstrap` option and prepend the `ssh2.exec` function by its content and `&&`.

Working proof of concept (sorry for the high-res):

![image](https://cloud.githubusercontent.com/assets/2091902/12960707/f322f316-d03c-11e5-81cf-e97f84833650.png)

WebFreak001:
Oh I thought that was some execute process and not spawn shell command. Shell would make everything easier then. Can you PR then?

bbenoist:
No problem, I'll open the PR.
Just give me enough time to document the new option properly :wink:

PR #16, opened 2016-02-10
Add an option to execute additional commands before GDB when using SSH

bbenoist:
See #15 for more info.


WebFreak001:
Please also add the new value with description as string to the package.json so it has auto complete

bbenoist:
:+1:

PR #25, opened 2016-02-22
Add the -f flag to break-insert

bbenoist:
To also support shared libraries which are loaded after the breakpoint
insertion.

Additional info about the issue can be seen [here (NetBeans bugtracker)](https://netbeans.org/bugzilla/show_bug.cgi?id=144897).


WebFreak001:
looks good, thanks

PR #81, opened 2016-11-14
Quote input when adding breakpoint

### LINKED ISSUE #80: Native debug on macOS is becoming inoperable with each VSCode update
The native debug extension is becoming unusable with each update of VS Code for macOS. The lldb-mi option to display the local variables capability vanished after 1.5. The gdb extension is not unusable anymore because it appears to ignore breakpoints. It would be nice to get this extension working as it was with VS Code 1.4.

rookwood101:
File path and line number are now double quoted when specified explicitly. Before only when using raw input was it double quoted.

Possibly fixes #80 and maybe others, however I was experiencing a problem with setting breakpoints on linux.

Also untested, would be great if a maintainer could - I don't have time to set up the development environment

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -468,7 +468,7 @@ export class MI2 extends EventEmitter implements IBackend {
 			if (breakpoint.raw)
 				location = '"' + escape(breakpoint.raw) + '"';
 			else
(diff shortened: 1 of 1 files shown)


rookwood101:
Good point. All okay now?

WebFreak001:
Looks good now thanks

PR #87, opened 2016-12-11
Add config option to pass arguments to GDB

Mrmaxmeier:
Hey,
this PR adds a config option that can be used to pass arguments to GDB.
I'm using it to disable my `.gdbinit` file by setting `gdbargs` to `["-nx", "-x ~/.custom_gdbinit"]` because my `.gdbinit` expects a real tty.
If you think a generic `debugger_args` option shared between all backends would be better though I'm open to push changes.

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -27,7 +27,7 @@ function couldBeOutput(line: string) {
 const trace = false;
 
 export class MI2 extends EventEmitter implements IBackend {
-	constructor(public application: string, public preargs: string[]) {
+	constructor(public application: string, public preargs: string[], public extraargs: string[]) {
 		super();
 	}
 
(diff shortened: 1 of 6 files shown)


src/backend/mi2/mi2.ts
```diff
@@ -27,7 +27,7 @@ function couldBeOutput(line: string) {
 const trace = false;
 
 export class MI2 extends EventEmitter implements IBackend {
-	constructor(public application: string, public preargs: string[]) {
+	constructor(public application: string, public preargs: string[], public extraargs: string[]) {
```
Mrmaxmeier: @WebFreak001 I've updated `MI2` to accept the additional arguments.

WebFreak001:
ok thanks for the addition. Also thanks for fixing the travis build issue, I wasn't sure what was causing it when I upgraded typescript and wanted to fix it only later because I thought it was a problem with travis.

PR #89, opened 2017-01-07
Removed outDir for outFiles

andyneff:


WebFreak001:
Uh does this fix anything or what is the goal of this?

andyneff:
Just future proofing. `outDir` is on the path to deprecation [apparently].(https://code.visualstudio.com/updates/v1_6#_glob-pattern-support-for-source-map-setup)

PR #90, opened 2017-01-07
WIP: Adding Docker features

andyneff:
My team is starting to use VS Code in Windows and Linux to debug Linux Containers in Docker. I am starting to add some **commands** for use in the `launch.json` files.

What we want to be able to do is:

- *Easily* attach to a process running in a container.
- Set breakpoints
- Read what line of code I'm on. All the other remote debugging solution I've looked out are missing the filename translation step needed for debugging over ssh, but code-debug works!
- Step through and in code (as long as I have symbols/source code, etc...)
- Detach and not have the pid die
- And of course be able to look at variables, etc...

From my initial ssh test, all of this seems possible!

Currently it will let you pick a container from the list of running containers, and then pick a pid of interest from the list of pids. Then the normal ssh feature works from there (maybe eventually I'll get `docker exec -it gdb` working, but for now **ssh works**, which is more than I can say on the microsoft [plugin](https://github.com/Microsoft/vscode-cpptools/issues/265#issuecomment-250229293))

TODO before I consider this no longer a WIP:

- ~~Pick container name~~ Done
- ~~Pick pid in container~~ Done
- Identify the tcp port exposed by docker
- Make it so when I disconnect the gdb detaches and ends.
- Auto generate ssh key via docker exec
- *Maybe* use direct gdb execution in docker as a remote debug (Not sure how to do this)
- Look into support gdbserver multi mode
- Look into supporting gdbserver to act more like ssh

I have to add all these docker feature to code-debug because it is the only way I know how to add command to the launch.json for use by the debugger. I tried adding as a separate extension, and I could not figure out how to get it working. So if there is something I'm missing, I'll gladly make this code a separate project.

Sample launch

```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Webfreak Debug2",
            "type": "gdb",
            "request": "attach",
            "dockerName": "${command.DockerName}",
            "target": "${command.DockerPid}",
            "cwd": "${workspaceRoot}",
            "ssh": {
                "forwardX11": false,
                "host": "localhost",
                "port": 32768,
                "cwd": "/opt/vip",
                "keyfile": "D:\\gdb_rsa",
                "user": "root"
            }
        }
    ]
}
```

Typescript is not my language of choice, so sorry for some crude code.

### DIFF
--- a/src/frontend/picker.ts
+++ b/src/frontend/picker.ts
@@ -1,0 +1,158 @@
+import * as vscode from "vscode";
+import * as child_process from "child_process";
+import * as os from "os";
+
+interface IItem {
+    name: string;
+    id: string; //This is what gets returned eventually
+}
+
+class DockerItem implements IItem{
+    name: string;
+    id: string;
+    protected image: string;
+    protected labels: string;
+
+    constructor(id, name, image, labels) {
+        this.name = name + "["+id+"]";
+        this.id = id;
+        this.image = image
+        this.labels = labels;
+    }
+    
+    toAttachItem() {
+        return {
+            label: this.name,
+            description: this.labels,
+            detail: this.image,
+            id: this.id
+        };
+    };
+};
+
+class ProcessItem implements IItem{
+    name: string;
+    id: string;
+    protected commandLine: string;
+
+    constructor(name, pid, commandLine) {
+        this.name = name;
+        this.id = pid;
+        this.commandLine = commandLine;
+    }
+
+    toAttachItem() {
+        return {
+            label: this.name,
+            description: this.id,
+            detail: this.commandLine,
+            id: this.id
+        };
+    };
+};
+
+abstract class Picker{
+    launchConfig: any;
+
+    getItems(){
+        return this.getEntries().then(function (processEntries: any[]) {
+            return processEntries.map(function (p) { return p.toAttachItem(); });
+        });
+    };
+
+    pickItem(launchConfig){
+        this.launchConfig = launchConfig;
+
+        var attachPickOptions = {
+            matchOnDescription: true,
+            matchOnDetail: true,
+            placeHolder: "Select the docker name to attach to"
+        };
+        return vscode.window.showQuickPick(this.getItems(), attachPickOptions)
+            .then(function (dockerItem: any) {
+                return dockerItem ? dockerItem.id : null;
+            });
+    };
+
+    parseItems(items: string){
+    //Parses a line of the output and returns array of objects
+        var lines = items.split('\n');
+        var entries = [];
+        for (var i = 1; i < lines.length; i++) {
+            var line = lines[i];
+            if (!line) {
+                continue;
+            }
+            var docker_1 = this.lineParser(line);
+            entries.push(docker_1);
+        }
+        return entries;
+    }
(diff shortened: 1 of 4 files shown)


WebFreak001:
wow, thanks for the great effort! I've never used docker, so I can't really make use of it, but I'm sure others can and probably will. As you aren't using Typescript that often, here is a tip to avoid many bugs:

```ts
myCallbackFunction(function () { this.a = 5; }); // bug, this will be something different
```

instead write this:

```ts
myCallbackFunction(() => { this.a = 5; }); // works
```

The `() => {}` syntax will rewrite all `this` variable accesses to `this_` and create that variable before the function call if it doesn't exist yet and if its used. So then you don't need to manually type it like here: https://github.com/WebFreak001/code-debug/pull/90/files#diff-eb8b3d435a8da4be9b6c7efc015dcbffR115

When you are done with all changes just use the built-in Typescript formatter (ctrl-shift-i). All my code uses it and it looks good with nearly no effort.

Thanks again for your contribution.

src/frontend/picker.ts
```diff
@@ -0,0 +1,158 @@
...
+                if (stderr && stderr.length > 0) {
+                    reject(new Error(stderr));
+                    return;
+                }
+                resolve(stdout);
+            });
+        });
+    };
+
+    abstract getEntries(): any; //any[]
```
WebFreak001: `Thenable<any[]>` is probably what you would want here

src/frontend/picker.ts
```diff
@@ -0,0 +1,158 @@
...
+export class DockerPicker extends Picker {
+    getEntries() {
+        var psCommand = ("docker ps --format {{.ID}}@{{.Names}}@{{.Image}}@{{.Labels}}");
+        var _this = this
+        return this.execChildProcess(psCommand, null).then(function (items: string) {
+            return _this.parseItems('\n'+items); //Add bogus header line
+        });
+    }
+
+    lineParser(line: string) {
```
WebFreak001: Maybe add an error check here in case something goes wrong? This _might_ contain an error message instead of a@b@c@d or it could break in the future or with an old version probably too. Just use `vscode.window.showErrorMessage` to notify the user instead of crashing.

src/frontend/picker.ts
```diff
@@ -0,0 +1,158 @@
...
+
+    getEntries() {
+        var _this = this;
+
+        if (this.launchConfig.dockerName === "null"){return;}
+
+        var dockerName = this.launchConfig.dockerName;
+
+        var commColumnTitle = Array(DockerPidPicker.secondColumnCharacters).join("a");
+        var psCommand = ("docker exec "+dockerName+" ps -axww -o pid=,comm=" + commColumnTitle + ",args=");
```
WebFreak001: Can dockerName or commColumnTitle contain spaces? If yes then you should probably add quotes around it or use the overload that takes an array of values. I prefer the latter more because you have total control over spawning the process and don't need to worry about spaces or having some weird shell that doesn't support this syntax for some reason

GitMensch:
@andyneff if this PR is a WIP please mark it as a draft and if possible adjust the conflicting package.json. Thank you!

Issue #97, opened 2017-03-05
Run gdb commands before connection


Gaswigue:
With Renesas RX gdb (embedded system - remote GdbServer). We need to provide gdb commands before to make the connection.

For example "-gdb-set non-stop on" and only after we can use "target extended-remote :*****".

I find a way to workaround this issue.

First remove this 3 lines of mi2.js (222-224) on 0.17.0
                this.sendCommand("gdb-set target-async on"),
                this.sendCommand("environment-directory \"" + escape(cwd) + "\""),
                this.sendCommand("target-select remote " + target)

And set all needed commands in "autorun" argument in launch.json.

A way to solve this issues is maybe to provide a new arguments as "autorun" but to be triggered before.

Otherwise, it's a really good piece of extension !

WebFreak001:
Hm making those commands more configurable sounds like a good idea

Gaswigue:
I can code this feature if needed. But how do you want I implement it ? the "before-connection-autorun" argument is good for you ?

WebFreak001:
Yes, as config value with the default of `["gdb-set target-async on", "environment-directory \"$(cwd)\"", "target-select remote $(target)"]`. For the variables just use a simple replace and only allow cwd and target for now. If you find other good candidates for variables, you are free to implement them as well though

Gaswigue:
Are ["gdb-set target-async on", "environment-directory \"$(cwd)\"", "target-select remote $(target)"] going to package.json for default value ?
Doing so the parser doesn't expand the variable $(cwd) and $(target).
What I did

"autorunBefore": {
      "type": "array",
       "description": "GDB commands to be run before connection",
      "default": ["gdb-set target-async on",
                       "environment-directory \"$(cwd)\"",
                       "target-select remote $(target)"
}

or should I check for ===undefined
and initiate the default value after ?

It's my first attemp to dev on vscode extension

WebFreak001:
if you include them in the package.json you need to escape the string like this I think: https://github.com/WebFreak001/code-debug/blob/master/package.json#L305

WebFreak001:
And yes you need them both in the package.json as well as in the js code by checking for undefined. The package.json default one is the auto completion the user will receive, your code needs to have a value that will replace the value

spir6s:
Are there any plans to integrate this into master?

dlshutch:
Just moving to vscode and am blown away by all the features.
This extension is excellent, and I appreciate all the work to make it happen!
[...]

PR #99, opened 2017-03-08
Run gdb commands before connection #97

Gaswigue:


WebFreak001:
It would be cool if environment-directory and target-select could also be inside the autorunBeforeCmds array. In line 232 where you only have sendCommand right now you would need to add `.replace(/\$escaped_cwd/g, escape(cwd)).replace(/\$cwd/g, cwd).replace(/target/g, target)` after the command string. This would make it much more configurable and it is really easily done like that. Then you can add `"environment-directory \"$escaped_cwd\""` and `"target-select remote $target"` to the autorunBeforeCmds array

Gaswigue:
It cross my mind to do as you say.
But as i am not very good at typescript/JS, I though there was a much better way to do so i didnt do it.
Thanks to your help. I'll check on that.

WebFreak001:
If you need any more help or suggestions, just push the code and I can use the github review feature. I think it's important to add this to SSH too, but you don't need to do that, that can be done in a separate PR or commit by me if you don't want to do it

Gaswigue:
Ssh is not needed for me but I will code the improvement for this use case also. I have some other ideas ;)

src/gdb.ts
```diff
@@ -42,9 +36,8 @@ class GDBDebugSession extends MI2DebugSession {
 		this.sendResponse(response);
 	}
 
-	protected launchRequest(response: DebugProtocol.LaunchResponse, args: LaunchRequestArguments): void {
-		this.miDebugger = new MI2(args.gdbpath || "gdb", ["-q", "--interpreter=mi2"], args.debugger_args);
-		this.initDebugger();
+
+	private initiliaseDefaultValueArgs(args: CommonRequestArguments, attach: boolean) {
```
Gaswigue: New function to initialize unseted variable.
This avoid the launchRequest/attachRequest copy/past

src/gdb.ts
```diff
@@ -68,11 +64,47 @@ class GDBDebugSession extends MI2DebugSession {
...
+			this.attached = false;
+			this.needContinue = true;
+			if (!hasAutorunBeforeArgs)
+				args.autorunBefore.push("target-select $target");
+		}
+		args.autorun = args.autorun.map(
+			s => {return escape(s);
+		});
+
+		args.autorunBefore = args.autorunBefore.map(
```
Gaswigue: `args.autorun = args.autorun.map(s => {return escape(s);});`
Resolve \ escape in $(workspaceroot) if we use it to load binary from gdb (command load)

WebFreak001: don't use escape for this, it will put a `\` before all other `\` and `"` characters, this will manipulate user input and it will break all commands using `"` or `\`. I'm not quite sure what you mean with the workspaceroot.

Gaswigue: I have a set of commands to send to gdb after connection
"load ${workspaceRoot}\\bin\\binary.x"
if i don't escape the path is wrong it remove all the \
the final command is
"load c:somedirAsomedirB\bin\binary.x" and not c:\somedirA\somedirB\bin\binary.x

Gaswigue: I try to do
load ".\\bin\\binary.x" and it work. no need to use workspaceRoot. I can delete if needed this part.

src/gdb.ts
WebFreak001: The other parts of the program spell it initialize instead of initialise, can you keep it consistent and change it to the american spelling?

GitMensch:
What is the state of this PR? Is it a draft? Is it abandoned?

GitMensch:
It's been a while...
I'd like to take that up on current master (ideally after #323 and #316 were merged as those change the initialization):

* adding the `initializeToDefault()`
* add the new setting (I'd prefer "beforeConnect" instead of `autorunBefore` though)
* "update" the async setting, see #297

PR #107, opened 2017-05-20
Gdb var objects

gentoo90:
Replace own value parser with [GDB's variable objects](https://sourceware.org/gdb/onlinedocs/gdb/GDB_002fMI-Variable-Objects.html) interface.
This improves variables view and allows to show values from custom pretty-printers loaded into GDB.

| |Before|After|
|---|---|---|
|C++|cpp_before|cpp_after|
|Rust|rust_before|rust_after|

Tested with `vscode-1.12.2` and `GDB-7.12` on Gentoo Linux

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,6 +101,16 @@
 								"description": "Additional arguments to pass to GDB",
 								"default": []
 							},
+							"valuesFormatting": {
+								"type": "string",
+								"description": "Set the way of showing variable values. 'disabled' - show value as is, 'parseText' - parse debuggers output text into structure, 'prettyPrinters' - enable debuggers custom pretty-printers if there are any",
+								"default": "parseText",
+								"enum": [
+									"disabled",
+									"parseText",
+									"prettyPrinters"
+								]
+							},
 							"printCalls": {
 								"type": "boolean",
 								"description": "Prints all GDB calls to the console",
@@ -191,6 +201,16 @@
 								"type": "boolean",
 								"description": "If true this will connect to a gdbserver instead of attaching to a PID",
 								"default": false
+							},
+							"valuesFormatting": {
+								"type": "string",
+								"description": "Set the way of showing variable values. 'disabled' - show value as is, 'parseText' - parse debuggers output text into structure, 'prettyPrinters' - enable debuggers custom pretty-printers if there are any",
+								"default": "parseText",
+								"enum": [
+									"disabled",
+									"parseText",
+									"prettyPrinters"
+								]
 							},
 							"printCalls": {
 								"type": "boolean",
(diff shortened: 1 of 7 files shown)


WebFreak001:
Wow thanks, I didn't know about that one, that surely will improve the quality of the variable output. Also I didn't know typescript supported that fancy async syntax there.

But I found an issue here where it tries to expand something on a null pointer (it thinks there is still data and showing the arrow but there isn't actually anything there): 

Currently trying to figure out how to best fix this.

WebFreak001:
hm the empty variable looks exactly the same as normal variables so I don't really know how to fix that, well maybe not really that big of an issue.

However arrays (length prefixed `length` + `ptr` objects) are now broken with this new code. I tried doing the simple thing by adding [0] to the variable name like `var1[0]` and also `var1.ptr[0]` but both didn't really work. Also strings would be nicer if they were directly inlined without the long pointer value before them. (doing all the tests for this in D)

gentoo90:
I've never used D. Can you post the piece of code which gives this?
Arrays in C++ and Rust work fine.

WebFreak001:
Here just some test code where both the issues can be observed if you just halt anywhere in the main:

```d
import std.stdio;

struct A
{
	string b;
	A* child;
}

void main()
{
	string str = "abc";
	writeln(str);
	A a;
	a.b = "def";
	writeln(a);
}
```

gentoo90:
var objects for local variables now have the same names as that variables? so adding to watchlist by right click should work better. But according to GDB's documentation expressions can't work with dynamic objects

>exp
The expression to be shown to the user by the front end to designate this child. For example this may be the name of a structure member.
For a dynamic varobj, this value cannot be used to form an expression. There is no way to do this at all with a dynamic varobj.
For C/C++ structures there are several pseudo children returned to designate access qualifiers. For these pseudo children exp is ‘public’, ‘private’, or ‘protected’. In this case the type and value are not present.
A dynamic varobj will not report the access qualifying pseudo-children, regardless of the language. This information is not available at all with a dynamic varobj.

gentoo90:
I added launch config parameter `valuesFormatting` with 3 possible values:
* `disabled` - use variable objects but don't enable pretty-printers (they can be buggy so it may be useful to disable them)
* `parseText` - use the current way and parse result of `data-evaluate-expression` with `expandValue()`
* `prettyPrinters` - use variable objects and enable pretty-printers

WebFreak001:
yeah thats perfect, thanks for this contribution. Code quality and your responsiveness were great. Gonna merge this and push a new version to the marketplace in a minute.

Also the main reason why I want it to work with D out of the box is that I listed this extension on the D wiki as debugger plus I always recommend it to D people using vscode and I am always mostly testing it with D.

Issue #113, opened 2017-06-26
Possible to use relative paths on remote.


xoviat:
First of all, thanks for providing such an excellent extension.

When I insert a breakpoint, code-debug executes the following command over ssh:

```
7-break-insert -f "/path/to/ssh/cwd/src/file.f:15"
```

What needs to happen is this:

```
7-break-insert -f "src/file.f:15"
```

My configuration is this (with nonrelevant information removed)

```
            "name": "Launch Remote",
            "type": "gdb",
            "request": "launch",
            "target": "./executable",
            "printCalls": true,
            "cwd": "${workspaceRoot}",
            "ssh": {
                "forwardX11": false,
                "host": "host",
                "cwd": "/path/to/ssh/cwd",
            }
```

WebFreak001:
have you tried using `"cwd": "."` in the ssh section?

xoviat:
I need to change to the correct directory, so I put the following in bootstrap

`cd /path/to/ssh/cwd`

Then:
```
1-gdb-set target-async on
2-environment-directory "."
3-file-exec-and-symbols "executable"
4-environment-cd "."
5-exec-arguments 
6-break-delete
7-break-insert -f "C:\\Program Files (x86)\\Microsoft VS Code/src/file.f:15"
Running executable
8-exec-run
-gdb-exit
```

WebFreak001:
so it works with a workaround, having a setting which allows you to do that would probably be better

xoviat:
No it does not work with a workaround. The following line is not correct:

```
7-break-insert -f "C:\\Program Files (x86)\\Microsoft VS Code/src/file.f:15"
```

PR #114, opened 2017-06-28
ssh: enable relative paths

xoviat:


xoviat:
fixes #113.

WebFreak001:
you also need to add the relative bool to the package.json and add a description so users will find it. Also you need to implement it for lldb because it has separate code from gdb for startup

PR #129, opened 2018-02-02
Support multi-threading

### LINKED ISSUE #36: support multithread debugging

LeszekSwirski:
Fixes #36. Not everything works perfectly, in particular non-varobj local don't have a well-defined thread, but it's a good base to build on. I'm not 100% sure about the logic behind VAR_HANDLES_START, so the current array size increase may be overkills

Also includes some odds-and-ends fixing parsing and reference printing.

### DIFF
--- a/src/backend/mi_parse.ts
+++ b/src/backend/mi_parse.ts
@@ -205,8 +205,10 @@ export function parseMI(output: string): MINode {
 		let oldContent = output;
 		let canBeValueList = output[0] == '[';
 		output = output.substr(1);
-		if (output[0] == '}' || output[0] == ']')
+		if (output[0] == '}' || output[0] == ']') {
+			output = output.substr(1); // ] or }
 			return [];
+		}
 		if (canBeValueList) {
 			let value = parseValue();
 			if (value) { // is value list
--- a/src/mibase.ts
+++ b/src/mibase.ts
@@ -260,6 +260,7 @@ export class MI2DebugSession extends DebugSession {
 	protected threadsRequest(response: DebugProtocol.ThreadsResponse): void {
 		if (!this.miDebugger) {
 			this.sendResponse(response);
+			return;
 		}
 		this.miDebugger.getThreads().then(
 			threads => {
--- a/src/mibase.ts
+++ b/src/mibase.ts
@@ -235,6 +235,8 @@ export class MI2DebugSession extends DebugSession {
 				Promise.all(all).then(brkpoints => {
 					let finalBrks = [];
 					brkpoints.forEach(brkp => {
+						// TODO: Currently all breakpoints returned are marked as verified,
+						// which leads to verified breakpoints on a broken lldb.
 						if (brkp[0])
 							finalBrks.push(new DebugAdapter.Breakpoint(true, brkp[1].line));
 					});
@@ -256,13 +258,24 @@ export class MI2DebugSession extends DebugSession {
 	}
 
 	protected threadsRequest(response: DebugProtocol.ThreadsResponse): void {
+		if (!this.miDebugger) {
+			this.sendResponse(response);
+		}
 		this.miDebugger.getThreads().then(
 			threads => {
 				response.body = {
 					threads: []
 				};
 				for (const thread of threads) {
-					response.body.threads.push(new Thread(thread.id, thread.id + ":" + thread.name));
+					let threadName = thread.name;
+					// TODO: Thread names are undefined on LLDB
+					if (threadName === undefined) {
+						threadName = thread.targetId;
+					}
+					if (threadName === undefined) {
+						threadName = "<unnamed>";
+					}
+					response.body.threads.push(new Thread(thread.id, thread.id + ":" + threadName));
 				}
 				this.sendResponse(response);
 			});
--- a/src/mibase.ts
+++ b/src/mibase.ts
@@ -1,4 +1,5 @@
-import { DebugSession, InitializedEvent, TerminatedEvent, StoppedEvent, OutputEvent, Thread, StackFrame, Scope, Source, Handles } from 'vscode-debugadapter';
+import * as DebugAdapter from 'vscode-debugadapter';
+import { DebugSession, InitializedEvent, TerminatedEvent, StoppedEvent, ThreadEvent, OutputEvent, Thread, StackFrame, Scope, Source, Handles } from 'vscode-debugadapter';
 import { DebugProtocol } from 'vscode-debugprotocol';
 import { Breakpoint, IBackend, Variable, VariableObject, ValuesFormattingMode, MIError } from './backend/backend';
 import { MINode } from './backend/mi_parse';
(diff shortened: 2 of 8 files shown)


src/backend/mi_parse.ts
```diff
@@ -205,8 +205,10 @@ export function parseMI(output: string): MINode {
 		let oldContent = output;
 		let canBeValueList = output[0] == '[';
 		output = output.substr(1);
-		if (output[0] == '}' || output[0] == ']')
+		if (output[0] == '}' || output[0] == ']') {
+			output = output.substr(1); // ] or }
 			return [];
+		}
```
WebFreak001: uh any reason for this change? Could you include some test what this is supposed to fix or what didn't work in the tests file please

LeszekSwirski: I don't remember the exact command that caused it, but it was to do with an mi command returning {} followed by other data, where the other data was getting lost

LeszekSwirski: I've added an (artificial) test, which I can confirm did fail before this patch.

WebFreak001:
ok while quickly going over the changes I didn't notice any major issues so I will do more extensive testing later with all supported debuggers (not sure if mago supports multithreading, but it's just a side product anyway to allow D people on windows to debug if they really want to).

I don't really know anything about async in js so I need to lookup if any thread safety issues can occur there or if it's just fancy promises.

Otherwise I don't know how I feel about all the missing semicolons and using `'` for strings longer than 1 character yet lol

But this seems like a good and useful PR, thanks

LeszekSwirski:
async in JS is effectively equivalent to promise `then` chains, there's still no actual parallelism or threading. I can fix the semicolons and `'`s.

I've only tested with gdb so YMMV, single-threaded functionality shouldn't change with other compilers though I think.

LeszekSwirski:
Gentle reminder.

WebFreak001:
sorry was kind of busy with other stuff and didn't really feel like reviewing something large, I can do it now

WebFreak001:
**GDB**
- [x] single threaded still works locally
- [x] multi threaded now works locally
- [x] single threaded still works over SSH
- [x] multi threaded now works over SSH
- [ ] single threaded still works over gdbserver
- [ ] multi threaded now works over gdbserver
- [ ] single threaded still works when attaching
- [ ] multi threaded now works when attaching
Exception:
```
TypeError: Cannot read property 'getThreads' of undefined
    at GDBDebugSession.threadsRequest (/tmp/code-debug/out/src/mibase.js:239:24)
    at GDBDebugSession.dispatchRequest (/tmp/code-debug/node_modules/vscode-debugadapter/lib/debugSession.js:398:22)
    at GDBDebugSession._handleData (/tmp/code-debug/node_modules/vscode-debugadapter/lib/protocol.js:97:38)
    at Socket.inStream.on (/tmp/code-debug/node_modules/vscode-debugadapter/lib/protocol.js:18:44)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:191:7)
    at readableAddChunk (_stream_readable.js:178:18)
    at Socket.Readable.push (_stream_readable.js:136:10)
    at Pipe.onread (net.js:560:20): Error: TypeError: Cannot read property 'getThreads' of undefined
    at GDBDebugSession.threadsRequest (/tmp/code-debug/out/src/mibase.js:239:24)
    at GDBDebugSession.dispatchRequest (/tmp/code-debug/node_modules/vscode-debugadapter/lib/debugSession.js:398:22)
    at GDBDebugSession._handleData (/tmp/code-debug/node_modules/vscode-debugadapter/lib/protocol.js:97:38)
    at Socket.inStream.on (/tmp/code-debug/node_modules/vscode-debugadapter/lib/protocol.js:18:44)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:191:7)
    at readableAddChunk (_stream_readable.js:178:18)
    at Socket.Readable.push (_stream_readable.js:136:10)
    at Pipe.onread (net.js:560:20)
    at file:///opt/visual-studio-code/resources/app/out/vs/workbench/workbench.main.js:2644:530
    at Object.v [as _notify] (file:///opt/visual-studio-code/resources/app/out/vs/workbench/workbench.main.js:140:302)
    at Object.enter (file:///opt/visual-studio-code/resources/app/out/vs/workbench/workbench.main.js:143:568)
    at n.Class.derive._oncancel._run (file:///opt/visual-studio-code/resources/app/out/vs/workbench/workbench.main.js:144:866)
    at n.Class.derive._oncancel._error (file:///opt/visual-studio-code/resources/app/out/vs/workbench/workbench.main.js:144:383)
    at file:///opt/visual-studio-code/resources/app/out/vs/workbench/workbench.main.js:2223:975
    at _.e.dispatch (file:///opt/visual-studio-code/resources/app/out/vs/workbench/workbench.main.js:2225:291)
    at _.e.handleData (file:///opt/visual-studio-code/resources/app/out/vs/workbench/workbench.main.js:2224:815)
    at Socket.<anonymous> (file:///opt/visual-studio-code/resources/app/out/vs/workbench/workbench.main.js:2223:835)
    at emitOne (events.js:96:13)
```

**LLDB**
- [x] single threaded still works locally
- [x] multi threaded now works locally.
minor issue: thread names say undefined 
- [ ] single threaded still works over SSH (breakpoints don't work)
- [ ] multi threaded now works over SSH (breakpoints don't work)
- [ ] single threaded still works when attaching
- [ ] multi threaded now works when attaching


**Mago-MI**
~- [ ] single threaded still works locally~ (can't test because I can't run a VM because I disabled AMD-V in my bios because it locks my CPU after some random amount of time)
I don't think testing anything else here makes sense, because it is very limited

But I supposed most of these issues were present already, I am going to check for completeness though. The only thing I see when testing here right now is therefore that undefined in the thread name of lldb.

WebFreak001:
ok so the Thread exception when it fails to connect is new, but gdbserver still doesn't work.

I did find one issue though: breakpoints show as verified in the broken lldb ssh connection now, they were unverified like this  before

WebFreak001:
I can merge this as it is now because it is a good start, but I would prefer it if there were some `// TODO` comments where I just commented what behaviour changed

LeszekSwirski:
Thanks for the review, I've added some defensive ifs and TODOs. Regarding the breakpoints being verified, I think there's some semantic difference between breakpoints being verified, and being enabled. Since I get the "done" response back from lldb, I assume they're verified as far as lldb is concerned (modulo lldb mi bugs), but they might not actually be enabled because of various reasons. Regardless, there's a new callback-based breakpoint API coming, so that might do a better job of maintaining breakpoint state.

src/mibase.ts
```diff
@@ -256,13 +258,24 @@ export class MI2DebugSession extends DebugSession {
 	}
 
 	protected threadsRequest(response: DebugProtocol.ThreadsResponse): void {
+		if (!this.miDebugger) {
+			this.sendResponse(response);
```
WebFreak001: hm this should also return so the next call doesn't crash

PR #133, opened 2018-02-21
Add multi-process support

LeszekSwirski:
Add support for debugging applications which fork, gated behind a new configuration flag. GDB only.

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -198,6 +198,16 @@ export class MI2 extends EventEmitter implements IBackend {
 			cmds.push(this.sendCommand("file-exec-and-symbols \"" + escape(target) + "\""));
 		if (this.prettyPrint)
 			cmds.push(this.sendCommand("enable-pretty-printing"));
+		if (this.multiProcess) {
+			cmds.push(
+			  this.sendCommand("gdb-set follow-fork-mode parent"),
+				this.sendCommand("gdb-set detach-on-fork off"),
+				this.sendCommand("gdb-set non-stop on"),
+				this.sendCommand("gdb-set schedule-multiple on"),
+
+				this.sendCommand("interpreter-exec console \"handle SIGSYS nostop noprint\"")
+			);
+		}
 
 		return cmds;
 	}
@@ -356,7 +366,7 @@ export class MI2 extends EventEmitter implements IBackend {
 									else if (reason == "exited-normally")
 										this.emit("exited-normally", parsed);
 									else if (reason == "exited") { // exit with error code != 0
-										this.log("stderr", "Program exited with code " + parsed.record("exit-code"));
+										this.log("stderr", "Inferior exited with code " + parsed.record("exit-code"));
 										this.emit("exited-normally", parsed);
 									}
 									else {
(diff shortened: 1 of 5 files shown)


src/backend/mi2/mi2.ts
```diff
@@ -198,6 +198,16 @@ export class MI2 extends EventEmitter implements IBackend {
 			cmds.push(this.sendCommand("file-exec-and-symbols \"" + escape(target) + "\""));
 		if (this.prettyPrint)
 			cmds.push(this.sendCommand("enable-pretty-printing"));
+		if (this.multiProcess) {
+			cmds.push(
+			  this.sendCommand("gdb-set follow-fork-mode parent"),
```
WebFreak001: you mixed indentation here

WebFreak001:
not sure if it happens because of this PR but the continue button doesn't show it anymore when it continues on the normal mode + the UI doesn't update when the location changes (you need to click it again to update), maybe because of the new thread names now?

This PR functionality seems to be broken for me too: if I step over the fork call it just continues running and never breaks again and doesn't show a callstack, if I continue over a fork it shows the PIDs on the callstack but it doesn't show it continuing and it never breaks again too

LeszekSwirski:
Yeah, I should have said that this is very WIP, very experimental, I more wanted to share the general idea. There's definitely more work to be done.

One of the big differences with the single process mode is that I use non-stop mode, i.e. only one thread is paused after a break point, not all of them. Would that explain the behaviour you're seeing? Could you share the test program?

WebFreak001:
ah yeah that could be the reason.

Here the source: (compile with dmd -g app.d)
```d
import std.stdio;
import core.thread;
import core.sys.posix.unistd;
import std.process;

struct S
{
	int a;
	long b;
	string c;
	T* f;
}

struct T
{
	byte d;
	char e;
	T* child;
}

void main(string[] args)
{
	writeln("Waiting 10s for attach");
	Thread.sleep(10.seconds);
	foreach (key, value; environment.toAA)
		writeln(key, " = ", value);
	stdout.flush();

	for (int i = 0; i < 3; i++)
	{
		writeln(i);
		stdout.flush();
		Thread.sleep(1.seconds);
	}

	if (args.length > 1 && args[1] == "fork")
		fork();
	else if (args.length == 1 || args[1] != "single")
		new Thread(&secondThread).start();

	write("No newline!");
	stdout.flush();
	int count = 5;

	for (; count < 20; count++)
	{
		S s;
		s.f = new T();
		s.f.child = new T();
		s.f.child.d = 4;
		s.a = count;
	}

	writeln("Got Arguments: ", args);
}

void secondThread()
{
	for (int i = 0; i < 10; i++)
	{
		writeln("Multithreaded ", i);
		stdout.flush();
		Thread.sleep(1.seconds);
	}
}
```

Execute it with arguments = `fork` in the debugger extension to make it fork instead of multithread, add some breakpoints here and there. Here is a binary if you are on linux: 
[app.zip](https://github.com/WebFreak001/code-debug/files/1745718/app.zip)

I mean that's not the way how you would call fork in an actual application but it does something which should be debuggable.

GitMensch:
@LeszekSwirski that's an interesting issue - can you update the conflicting files and, if you still see it as a WIP mark it as draft?

LeszekSwirski:
I've rebased the PR and marked it as draft, I don't currently have the cycles to investigate this any further though.

GitMensch:
Thank you @LeszekSwirski for the update, so it is now up to @WebFreak001 or others to take it up.

Issue #134, opened 2018-03-08
support passphrase encrypted ssh keys


JelleRoets:
My default key ~/.ssh/id_rsa is passphrase encrypted, however I stored the passphrase in the keychain. (I'm using OSX 10.13). So I can just login to my remote using `ssh [username]@[address]` without having to specify the passphrase every time.

When I specify a keyfile in the ssh config object of a debug launch, I get the following error:
"Failed to SSH: Error: Encrypted private key detected, but no passphrase given"
![image](https://user-images.githubusercontent.com/4798064/37175658-7d79440c-231a-11e8-8ffb-cd80f09fb745.png)

JelleRoets:
Solved by the above merged PR.
Solution: if you want to automatically use a passphrase protected ssh key:
1. first add your key to the ssh agent and store the passphrase, as described in https://www.ssh.com/ssh/add 
2. In the launch.json add a property "useAgent": true to the "ssh" object, and remove any "password" or "keyfile" objects

PR #136, opened 2018-03-17
add option to use ssh agent for authentication

JelleRoets:
Added an option "useAgent" to the "ssh" launch arguments. When this boolean is set to true, it automatically uses the active ssh agent (via SSH_AUTH_SOCK environment variable) for authentication  instead of the other options: "password" or "keyfile".

This should also solve the problem of trying to use a private ssh keyfile that is protected by a passphrase. With this approach the agent will automatically use the default ssh key and lookup the passphrase in the keychain (when it is stored there), this prevents the unsafe requirement of specifying the ssh password or passphrase as plane text in the launch.json file. #134 

(Only tested on Mac OSX 10.13)

### DIFF
--- a/package.json
+++ b/package.json
@@ -162,6 +162,11 @@
 										"type": "string",
 										"description": "Absolute path to private key"
 									},
+									"useAgent": {
+										"type": "boolean",
+										"description": "Auto-detect the running SSH agent (via SSH_AUTH_SOCK environment variable) and use it to perform authentication",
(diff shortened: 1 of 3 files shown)


WebFreak001:
oh wow surprisingly easy

WebFreak001:
will release as soon as I have my private key back because my hard drive died and now I need to extract an old backup and hope the key is in there

PR #138, opened 2018-03-20
Add explicit fortran language Ids

rafmudaf:
This pull request adds the explicit language IDs for fortran files from the Visual Studio Code plugin [Modern Fortran](https://github.com/krvajal/vscode-fortran-support/blob/master/package.json).

Without these IDs, breakpoints cannot be added to at least `.f90` and `.F90` files.


WebFreak001:
shouldn't there rather be an issue opened at fortran for not using `fortran90` or vscode for not respecting aliases? Including this there are 5 fortrans in the config now. Plus you only added this to gdb and not to lldb too (can you copy the other fortrans from there over to lldb too?)

rafmudaf:
There should be an issue opened at fortran or vscode but its not entirely clear what and where the problem is. This solves **my** problem, not **the** problem, but free time to solve more problems is hard to find.

lldb hasn't supported fortran for a few years. Are you trying to keep that language listed under lldb for older version compatibility?

WebFreak001:
uh just keeping both as they usually have a similar featureset, lldb already has one fortran in there so it should contain all fortrans we add to gdb

rafmudaf:
@WebFreak001 ok I've updated this pull request to include all the fortran language id's in both gdb and lldb languageId lists

PR #149, opened 2018-07-04
add valuesFormatting by default

### LINKED ISSUE #148: Complex data types aren't supported
If submitting a bug please make sure

- [x] If you are using gdb
  - [x] `gdb --version` >= 7.7.1
  - [x] it works on the command line with `gdb`
  - [x] `cwd` and `target` are properly set

here is a list of local variables:

image

Here is a code: `std::vector<std::string> aa{"a", "bbb", ""};`

As you see `aa` isn't pretty expanded

WebFreak001:
by @Yanpas fix #148

### DIFF
--- a/package.json
+++ b/package.json
@@ -346,7 +346,8 @@
 							"request": "launch",
 							"name": "${2:Launch Program}",
 							"target": "${1:./bin/executable}",
-							"cwd": "^\"\\${workspaceRoot}\""
+							"cwd": "^\"\\${workspaceRoot}\"",
+							"valuesFormatting": "parseText"
 						}
 					},
 					{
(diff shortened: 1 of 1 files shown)


WebFreak001:
@Yanpas can you add it to all `configurationSnippets` and `initialConfigurations` blocks and not just to those 2 specific GDB ones?

Yanpas:
I've amended commit

WebFreak001:
can you also add it to the `initialConfigurations` blocks? These are the default configurations you get if you just hit that gear and select "Debug (GDB)" without any existing debug config.

Yanpas:
done

PR #150, opened 2018-07-17
Enable tslint with basic rules

simark:
Add tslint as a plugin to tsc, so that they will show up in vscode or
whatever editor people use.  I only enabled two basic rules that I think
are very useful, prefer-const and no-var-keyword.  Violations of these
rules were fixed with "tslint -p . --fix".  There was one unused
variable (in the isExpandable) function that I removed by hand.


simark:
Context: We are looking into adding C/C++ debug support using GDB in the [Theia](https://github.com/theia-ide/theia/) project.  We are already planning on using the Debug Adapter Protocol for other debuggers.  Instead of starting from scratch for C/C++, I am considering using code-debug (and therefore probably contributing to it).  Since I really appreciate what tslint does when writing code for Theia, I thought I would test the water by contributing this simple patch.

WebFreak001:
hm why no-var-keyword? Isn't var just the same as let...

simark:
> hm why no-var-keyword? Isn't var just the same as let...

No, var is there only for compatibility with Javascript (so that Javascript code is also valid Typescript code).  But the behavior is different regarding to scopes, `let` doing the thing you would expect from any other language.

See this for example: https://basarat.gitbooks.io/typescript/content/docs/let.html

src/backend/gdb_expansion.ts
```diff
@@ -51,16 +50,16 @@ export function expandValue(variableCreate: Function, value: string, root: strin
...
 			stringEnd++;
 		}
-		let str = value.substr(0, stringEnd).trim();
+		const str = value.substr(0, stringEnd).trim();
 		value = value.substr(stringEnd).trim();
 		return str;
 	};
 
-	let stack = [root];
+	const stack = [root];
```
WebFreak001: uh this variable gets changed (in line 65)

WebFreak001:
```js
var foo = 123;
if (true) {
    var foo = 456;
}
console.log(foo); // 456
```
wtf js...

WebFreak001:
cool thanks

PR #151, opened 2018-07-18
Some more tslint rules

simark:
I enabled a few more rules.  They were mostly fixed tslint's `--fix`.

Some of them are about formatting.  I don't mind what the formatting rules are, but I think it helps if they are enforced by the linter, so you don't loose time later arguing/commenting about spaces/semicolons/etc.


WebFreak001:
hm ok cool, didn't think there were so many missing semicolons. Definitely good to have it all consistent now though.

PR #177, opened 2019-05-07
add gdb goto support

ntoskrnl7:
Add support for set next statement(goto request). GDB, LLDB support


WebFreak001:
sorry I forgot about this PR.

I added filename quoting (for files with spaces) and will merge this as it works as expected

PR #188, opened 2019-08-03
add zig support

evangrayk:
This extension seems to already support zig debugging, all that's needed is to support the file type and use the [zig extension](https://github.com/ziglang/vscode-zig).

See also [this guide to debugging zig with this extension](https://www.reddit.com/r/Zig/comments/cl0x6k/debugging_zig_in_vscode/)

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,8 @@
 						"asm",
 						"vala",
 						"crystal",
-						"kotlin"
+						"kotlin",
+						"zig"
 					]
 				},
 				"variables": {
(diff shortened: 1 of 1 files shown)


WebFreak001:
is it only supported with GDB and not with LLDB?

evangrayk:
I tested this with `lldb-mi` on mac where it seems to work, and [here](https://www.reddit.com/r/Zig/comments/cl0x6k/debugging_zig_in_vscode/) it sounds like it works with gdb as well.

WebFreak001:
can you also add zig to the lldb enabled languages part then?

PR #210, opened 2020-03-03
Display thread details if thread name is missing

anshulrouthu:
thread name information is present in thread details column in gdb server.  Use this details information is name is missing.

### DIFF
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -15,6 +15,7 @@ export interface Thread {
 	id: number;
 	targetId: string;
 	name?: string;
+	details?: string;
 }
 
 export interface Stack {
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -612,6 +612,11 @@ export class MI2 extends EventEmitter implements IBackend {
 				ret.name = name;
 			}
 
+			const details = MINode.valueOf(element, "details");
+			if (details) {
+				ret.details = details;
+			}
+
 			return ret;
 		});
 	}
--- a/src/mibase.ts
+++ b/src/mibase.ts
@@ -277,6 +277,9 @@ export class MI2DebugSession extends DebugSession {
 					let threadName = thread.name;
 					// TODO: Thread names are undefined on LLDB
 					if (threadName === undefined) {
+						threadName = thread.details;
+					}
+					if (threadName === undefined) {
 						threadName = thread.targetId;
 					}
 					if (threadName === undefined) {


anshulrouthu:
I see this in the jlink gdb server from segger, JLinkGDBServer.exe. I am
not sure if there is a better way to see the thread name. The target is
cortex-m4 processor.

WebFreak001:
can you maybe send a small log dump of how the threads are sent back via GDB MI?

anshulrouthu:
This is the output of "info threads" command in gdb console. I had to mask
the file paths, because it is a proprietary software or else I am going to
be in trouble.See after the target id the next field is just blank string
"" which is what the Native Debug extension is trying to parse but its
empty, the next column is the details field which contain the name of the
thread, I think this might only be specific to the GDBServerRTOS from
segger.

  Id   Target Id         Frame
* 2    Thread 536772672 "" (HWM main) ASSERT_TRAP () at
..\..\..\***\******\***_assert.c:493
  3    Thread 537091168 "" (MNAND MEM 0) HWM_sgnl_tsk_switch
(i_list_of_cpus_to_signal=i_list_of_cpus_to_signal@entry=0) at
..\..\..\***********\*******\***_cortex.c:118
  4    Thread 536736128 "" (hwm io main) HWM_sgnl_tsk_switch
(i_list_of_cpus_to_signal=i_list_of_cpus_to_signal@entry=0) at
..\..\..\*********\************\***_cortex.c:118
  5    Thread 536699512 "" (TSK 0bkgd) hwm_init_done () at
..\..\..\***\***_main.c:1556

Here is the trace from the plugin

stdout: 30^done,threads=[{id="2",target-id="Thread 536772672",details="HWM
main",name="",frame={level="0",addr="0x0000361a",func="ASSERT_TRAP",args=[],file=".
.\\..\\..\\******\\****_assert.c",fullname="D:\\********\\****_assert.c",line="493"},state="stopped"},{id="3",target-id="Thread
537091168",details="MNAND MEM
0",name="",frame={level="0",addr="0x00026e04",func="HWM_sgnl_tsk_switch",args=[{name="i_list_of_cpus_to_signal",value="0"},{name="i_list_of_cpus_to_signal@entry
",value="0"}],file="..\\..\\..\\***\\**********\\cortexm\\*****_cortex.c
",fullname="D:\\************\\processor\\cortexm\\****_cortex.c",line="118"},state="stopped"},{id="4",target-id="Thread
536736128",details="hwm io
main",name="",frame={level="0",addr="0x00026e04",func="HWM_sgnl_tsk_switch",args=[{name="i_list_of_cpus_to_signal",value="0"},{name="i_list_of_cpus_to_signal@entry
",value="0"}],file="..\\..\\..\\********\\processor\\cortexm\\***_cortex.c
",fullname="D:\\**********\\processor\\cortexm\\***_cortex.c",line="118"},state="stopped"},{id="5",target-id="Thread
536699512",details="TSK
0bkgd",name="",frame={level="0",addr="0x00030714",func="hwm_init_done",args=[],file=".
.\\..\\..\\****\\****_main.c",fullname="D:\\*************\\***_main.c
",line="1557"},state="stopped"}],current-thread-id="2"

Anshul

WebFreak001:
thanks! I included your change but made some other modifications to it, the name is determined on the mibase side and no longer in the debugger side and I also removed name as usage. It now uses the target-id which is a better way than repeating the program name 20 times.

PR #221, opened 2020-05-01
add support for gdb path substitution

karljs:
When debugging a binary, if the build environment is different from the debugging environment (e.g., built in Docker, debugging locally) you might see errors where the debugger isn't able to find the source files that match up with the binary.  This manifests in the extension as error messages in the debug console along the lines of "No source file named ...".

GDB solves this problem with "path substitutions" in a variable.  So if you build a binary in `/root` and want to debug it in `/home/username` you need to issue the command `set substitute-path /root /home/username`. 

This PR adds a configuration option for `launch.json` that lets you set this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
 							},
 							"pathSubstitutions": {
 								"type": "object",
-								"description": "Help GDB find your source using path substitutions (GDB `substitute-path` variable",
+								"description": "Help GDB find your source using path substitutions (GDB `substitute-path)` variable",
 								"default": {
 									"<fromPath>": "<toPath>"
 								}
@@ -261,7 +261,7 @@
 							},
 							"pathSubstitutions": {
 								"type": "object",
-								"description": "Help GDB find your source using path substitutions (GDB `substitute-path` variable",
+								"description": "Help GDB find your source using path substitutions (GDB `substitute-path)` variable",
 								"default": {
 									"<fromPath>": "<toPath>"
 								}
@@ -531,7 +531,7 @@
 							},
 							"pathSubstitutions": {
 								"type": "object",
-								"description": "Help LLDB find your source using path substitutions (LLDB `target.source-map` variable",
+								"description": "Help LLDB find your source using path substitutions (LLDB `target.source-map)` variable",
 								"default": {
 									"<fromPath>": "<toPath>"
 								}
@@ -678,7 +678,7 @@
 							},
 							"pathSubstitutions": {
 								"type": "object",
-								"description": "Help LLDB find your source using path substitutions (LLDB `target.source-map` variable",
+								"description": "Help LLDB find your source using path substitutions (LLDB `target.source-map)` variable",
 								"default": {
 									"<fromPath>": "<toPath>"
 								}
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
 		"native",
 		"debug"
 	],
-	"version": "0.24.0",
+	"version": "0.25.0",
 	"publisher": "webfreak",
 	"icon": "images/icon.png",
 	"engines": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
 		"native",
 		"debug"
 	],
-	"version": "0.24.0",
+	"version": "0.25.0",
```
WebFreak001, line 12: please don't change the version property (I change it on release when I call vsce and make it a separate commit)

src/backend/mi2/mi2.ts
```diff
@@ -27,7 +27,7 @@ function couldBeOutput(line: string) {
 const trace = false;
 
 export class MI2 extends EventEmitter implements IBackend {
-	constructor(public application: string, public preargs: string[], public extraargs: string[], procEnv: any) {
+	constructor(public application: string, public preargs: string[], public extraargs: string[], procEnv: any, public pathSubstitutions: any = {}) {
```
WebFreak001, line 30: to avoid bloating this method any more in future changes, can you make this a `public extraCommands: string[] = []`?

src/backend/mi2/mi2.ts
```diff
@@ -197,6 +197,10 @@ export class MI2 extends EventEmitter implements IBackend {
 			cmds.push(this.sendCommand("file-exec-and-symbols \"" + escape(target) + "\""));
 		if (this.prettyPrint)
 			cmds.push(this.sendCommand("enable-pretty-printing"));
+		
+		for (const key in this.pathSubstitutions) {
+			cmds.push(this.sendCommand("gdb-set substitute-path " + key + " " + this.pathSubstitutions[key]));
+		};
```
WebFreak001, line 203: this logic would then be moved into GDB, or if this also works with LLDB put this into a function in MI2 which then modifies `this.extraCommands` and is called in gdb.ts before initDebugger

Instead it would here just do
```ts
this.extraCommands.forEach((cmd) => {
	cmds.push(this.sendCommand(cmd));
});
```

package.json
```diff
@@ -107,6 +107,13 @@
 								"description": "Additional arguments to pass to GDB",
 								"default": []
 							},
+							"pathSubstitutions": {
+								"type": "object",
+								"description": "Help GDB find your source using path substitutions (GDB `substitute-path` variable",
+								"default": {
+									"<fromPath>": "<toPath>"
+								}
+							},
```
WebFreak001, line 116: you need to copy this to the attach arguments too

src/gdb.ts
```diff
@@ -10,6 +10,7 @@ export interface LaunchRequestArguments extends DebugProtocol.LaunchRequestArgum
 	gdbpath: string;
 	env: any;
 	debugger_args: string[];
+	pathSubstitutions: any;
```
WebFreak001, line 13: make this a `{[index: string]: string}` type

package.json
karljs: Reverted that, sorry.

karljs:
Thanks for the suggestions.  I've made them all, and also added support for LLDB (with the caveat that I don't currently have it set up and can't test it thoroughly.)

package.json
```diff
@@ -107,6 +107,13 @@
 								"description": "Additional arguments to pass to GDB",
 								"default": []
 							},
+							"pathSubstitutions": {
+								"type": "object",
+								"description": "Help GDB find your source using path substitutions (GDB `substitute-path` variable",
```
WebFreak001, line 112: you didn't close the parentheses

package.json
```diff
@@ -252,6 +259,13 @@
...
```
WebFreak001, line 264: ditto

package.json
```diff
@@ -515,6 +529,13 @@
 								"description": "Additional arguments to pass to LLDB",
 								"default": []
 							},
+							"pathSubstitutions": {
+								"type": "object",
+								"description": "Help LLDB find your source using path substitutions (LLDB `target.source-map` variable",
```
WebFreak001, line 534: ditto

package.json
```diff
@@ -655,6 +676,13 @@
...
```
WebFreak001, line 681: ditto

karljs:
Good catch, thanks.  Fixed the typos

Issue #192, opened 2019-09-26
Missing support for remote GDB via intermediate remote server


illegal-instruction:
I'm trying to use the extension to debug some C code in a remote server. That server is defined in my .ssh/config as 'gdb_server' and is accessed via an intermediate remote server:

```
Host gdb_server
 Hostname xxx.yyy.zz.aa
 Port 2038
 User <someuser>
 ServerAliveInterval 60
 ProxyCommand ssh <intermediate_user>@<intermediate.server.com> nc %h %p %r

```
If I use this configuration:

```
 "ssh": {
   "host": "gdb_server",
   "cwd": "<app_root_path@remote_server>",
   "keyfile": "/home/(omitted)/.ssh/id_rsa",
    "user": "<intermediate_user>",
    "port": 2038,
    "bootstrap": "<some scripts>"
}
```
I get the following error:

```
Could not run gdb over ssh!
Error: getaddrinfo ENOTFOUND aurora gdb_server:2038
```
Using the 'port' to '22' returns the same error.

Couldn't find any other parameter to configure this kind of 'intermediate-ssh' setup. In other cases, I've been able to use the extension to debug in a server accessible from my computer without any issue.

I guess the idea should be (I have no idea how the extension works internally so I don't know how hard it is), to enable the support of 'ssh/config named configurations', like gdb_server in this example.

Thanks for your time,
C

P.D: In any case this extension is awesome!

WebFreak001:
the extension is using the https://www.npmjs.com/package/ssh2 library which is a pure JS implementation of SSH and doens't support loading that ssh/config file. So instead it would be necessary to load and parse the ssh config for the current user with all ssh options which we want to support considered.

mohamedchebbii:
@WebFreak001 : Can you please add the socksv5 to support remote GDB via intermediate remote server ?
as described here : https://www.npmjs.com/package/ssh2#dynamic-11-port-forwarding-using-a-socksv5-proxy-using-socksv5-

mohamedchebbii:
I changed the code, instead of using ssh, I used ssh-promise, and It's working

illegal-instruction:
@mohamedchebbii That is great, thanks a lot for your work. It would be possible for you to share your modified code or create a pull request that could help @WebFreak001 bringing this, if possible, into the extension?

Thanks both of you for your work!

PR #223, opened 2020-05-10
Missing support for remote GDB via intermediate remote server #192

mohamedchebbii:
Hello @WebFreak001,
thank you for your comments, please bear with me, I'm not familiar with npm packages.

I checked the ssh2 package: https://www.npmjs.com/package/ssh2#dynamic-11-port-forwarding-using-a-socksv5-proxy-using-socksv5
It seems that It support port forwarding, but I can't find where to add the username/password/ssh_key_file
but ssh2_promise do that: https://www.npmjs.com/package/ssh2-promise#connect-to-ssh-server-via-hopping

I tried in this pull request to change the minimum of code and I just remplaced ssh2 by ssh2-promise

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,14 @@
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
+		"@heroku/socksv5": {
+			"version": "0.0.9",
+			"resolved": "https://registry.npmjs.org/@heroku/socksv5/-/socksv5-0.0.9.tgz",
+			"integrity": "sha1-ejkFkhE2smZpeaD4a7TwYvZX95M=",
+			"requires": {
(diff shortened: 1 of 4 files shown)


GitMensch:
@mohamedchebbii I'm very interested in this, can you please take that up again from current master with plain ssh2?
The way to do the hoping is now explicit documented at https://www.npmjs.com/package/ssh2#connection-hopping, each parameters are "normally" passed on `connect`.

Maybe you can adjust the proxy settings to be wrapped in an array, this way one can setup zero (array `proxy === undefined`), one, or even 48 proxies (which would be quite bad, but still).
Even if only "at most one proxy" is supported for now the settings are already setup to add more later - without breaking existing launch configurations.

PR #242, opened 2020-09-14
Add a register view

nomtats:
Hi,

I've added a register view using a TreeView.
It doesn't distinguish GPR from the other types of registers at all.
Let me know if you think this is acceptable.
image

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -1,4 +1,4 @@
-import { Breakpoint, IBackend, Thread, Stack, SSHArguments, Variable, VariableObject, MIError } from "../backend";
+import { Breakpoint, IBackend, Thread, Stack, SSHArguments, Variable, RegisterValue, VariableObject, MIError } from "../backend";
 import * as ChildProcess from "child_process";
 import { EventEmitter } from "events";
 import { parseMI, MINode } from '../mi_parse';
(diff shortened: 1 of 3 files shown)


GitMensch:
That looks like a neat idea. Checking the discussion within multiple vscode extensions the separate register panel was "dropped"; instead the "way to go" is using the Variable panel.

For "old" vscode versions this means to add a "virtual" variable "Registers" with those beneath the local variables; newer versions (not checked which one exactly) can return those with "Register" scope and vscode will do the right thing.

Could you please inspect to:

* Add the registers to the dap implementation? That was [added in 2019](https://github.com/microsoft/debug-adapter-protocol/pull/49), so should be possible.
* If the registers aren't shown because of too low dependency version: recheck which version is needed and possibly add a "virtual" variable that contains those.

jbreu:
Any chance this gets merged soon? It would be really helpful!

WebFreak001:
> Any chance this gets merged soon?

probably not unless someone puts in the work to bring it up to date with the current extension and vscode capabilities. If you want to give it a try, we can help you if you run into issues.

you can find a guide on working on the project here: https://github.com/WebFreak001/code-debug/blob/master/HACKING.md (it's even fairly recent, thanks @GitMensch!)

jbreu:
Grüß Gott,
I will have a look, but I have no clue about these Typescript and npm workflows so I wouldnt count on me ;)

nomtats:
Hi @WebFreak001,

I've found some time to update this to work with the latest DAP.
Let me know if this works for you.

Screenshot 2024-02-07 at 12 53 31

src/mibase.ts
```diff
@@ -436,64 +437,74 @@ export class MI2DebugSession extends DebugSession {
...
-			let stack: Variable[];
 			try {
-				stack = await this.miDebugger.getStackVariables(id.threadId, id.level);
-				for (const variable of stack) {
-					if (this.useVarObjects) {
-						try {
-							const varObjName = VariableScope.variableName(args.variablesReference, variable.name);
-							let varObj: VariableObject;
+				if (id.name != "Registers") {
+					const stack: Variable[] = await this.miDebugger.getRegisters();
```
WebFreak001, line 442: why is this using `getRegisters` inside `id.name != "Registers"`?

src/mibase.ts
```diff
@@ -436,64 +437,74 @@ export class MI2DebugSession extends DebugSession {
...
 		if (id instanceof VariableScope) {
-			let stack: Variable[];
 			try {
-				stack = await this.miDebugger.getStackVariables(id.threadId, id.level);
-				for (const variable of stack) {
-					if (this.useVarObjects) {
-						try {
-							const varObjName = VariableScope.variableName(args.variablesReference, variable.name);
-							let varObj: VariableObject;
+				if (id.name != "Registers") {
```
WebFreak001, line 441: consider making this `id.name == "Registers"` and swapping around the if and else blocks, since that's more readable and more extensible if we add more views

src/backend/mi2/mi2.ts
```diff
@@ -758,6 +758,57 @@ export class MI2 extends EventEmitter implements IBackend {
 		return ret;
 	}
 
+	async getRegisters(): Promise<Variable[]> {
+		if (trace)
+			this.log("stderr", "getStackRegisters");
```
WebFreak001, line 763:
```suggestion
			this.log("stderr", "getRegisters");
```

src/mibase.ts
nomtats: Fixed

src/backend/mi2/mi2.ts
nomtats: Fixed

src/mibase.ts
nomtats: Fixed.

WebFreak001 (approved):

WebFreak001:
thanks for tackling this again!

jbreu:
Thank you very much @nomtats !

jbreu:
@nomtats @WebFreak001 one question, if I may: How do you manage to get always hex output? For me its a mix of decimal and hex when setting "valuesFormatting": "prettyPrinters" in launch.json

grafik

nomtats:
@jbreu it uses 'data-list-register-values N' to retrieve register values where 'N' means "natural format". In another word, it's whatever GDB feels like.
https://ftp.gnu.org/old-gnu/Manuals/gdb/html_node/gdb_223.html
you can change it to 'x' to always get Hex values.

Zapmawww:
hi guys, what about add `--skip-unavailable` to the the gdb command, so that we could ignore unavailable regs and still get a output.
like this: `-data-list-register-values N --skip-unavailable`

Issue #206, opened 2020-02-04
'debugger_args' not honored for 'attach' configuration in gdb


rbrodeur:
Additional arguments to gdb in 'debugger_args' are not being passed to the gdb command line when using the 'attach' connection method. In the embedded environment where I work, the '-x gdb.ini' parameter is required to set up gdb correctly to attach to the target. The 'autorun' configuration seems to happen too late and is not working in my environment.

It would be super helpful if the 'attach' connection method concatenated the 'debugger_args' the same way as the launch connection method seems to (note, I have not validated that the 'launch' method actually correctly concatenates 'debugger_args', I'm just looking at the code differences between launch and attach).

load():
```
const args = this.preargs.concat(this.extraargs || []);
```

attach():
```
if (target.startsWith("extended-remote")) {
    isExtendedRemote = true;
    args = this.preargs;
} else
    args = args.concat([executable, target], this.preargs);
```

Thank you for taking the time to create and maintain this vscode extension and thank you for your consideration.

florin-saftoiu:
Same problem with the connect() method

PR #253, opened 2020-12-31
Fixed extraargs being ignored

mariotaku:
I noticed `debugger_args` option was ignored while attaching to remote gdbserver. This PR should fix this issue.


GitMensch:
related to #206, superseded by #316
@mariotaku it would be nice if you could review that

PR #259, opened 2021-03-01
Clear breakpoints only within the same file.

### LINKED ISSUE #230: Breakpionts lost.
I am using [rr](https://rr-project.org/) as a gdb backend.

$ rr --version
rr version 5.3.0

Config:
```json
      {
        "type": "gdb",
        "request": "attach",
        "name": "Attach to gdbserver",
        "executable": "${workspaceFolder}/x264",
        "target": ":50505",
        "remote": true,
        "cwd": "${workspaceRoot}",
        "valuesFormatting": "parseText"
      },
```

Steps to reproduce:
Step 1: `rr replay x264-2 -s 50505 -k`
Step 3: Attach to rr and add some breakpionts. It works fine.
Step 4: Disabling any breakpoint causes all breakpoints lost.

code-debug-rr

kvinwang:
Trying to fix #230.

When user set/remove a breakpoint, code-debug will remove all breakpoints and then add back the ones within the same file. That cause the breakpoints in other files lost.
This patch is trying to fix it by removing only the ones within the same file.

### DIFF
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -64,7 +64,7 @@ export interface IBackend {
 	loadBreakPoints(breakpoints: Breakpoint[]): Thenable<[boolean, Breakpoint][]>;
 	addBreakPoint(breakpoint: Breakpoint): Thenable<[boolean, Breakpoint]>;
 	removeBreakPoint(breakpoint: Breakpoint): Thenable<boolean>;
-	clearBreakPoints(): Thenable<any>;
+	clearBreakPoints(source?: string): Thenable<any>;
 	getThreads(): Thenable<Thread[]>;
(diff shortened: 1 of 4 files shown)


GitMensch:
This looks reasonable to me, @brownts @WebFreak001 do you see any reason to _not_ pull it in?

brownts:
Looks reasonable to me as well.

GitMensch:
While I did not tested it @kvinwang obviously did, and we had an additional positive review - and it actually looks better to me then the original, too - so merged.
Thanks for your contribution!

Note: I'd still would like to see a refactoring of its use - see https://github.com/WebFreak001/code-debug/pull/306#issuecomment-981961534 - but that's a separate thing and _possibly_ also a pure personal opinion.

PR #272, opened 2021-06-13
allow untrusted workspaces to be debugged

GitMensch:
Per "guidance for debug extensions" as noted in the [Workspace Trust Extension Guide](https://github.com/microsoft/vscode/issues/120251#issuecomment-825832603).

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
 		"debug"
 	],
 	"license": "public domain",
-	"version": "0.25.0",
+	"version": "0.25.1",
 	"publisher": "webfreak",
 	"icon": "images/icon.png",
 	"engines": {
--- a/package.json
+++ b/package.json
@@ -29,6 +29,11 @@
 		"type": "git",
 		"url": "https://github.com/WebFreak001/code-debug.git"
 	},
+	"capabilities": {
+		"untrustedWorkspaces": {
+			"supported": true
+		}
+	},
 	"contributes": {
 		"commands": [
 			{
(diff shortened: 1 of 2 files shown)


GitMensch:
Rechecked - no need to adjust the dependency version, just add the `cababilities.untrustedWorkspaces` and everything is fine.

WebFreak001:
this allows arbitrary code execution because the launch.json can input any executable path for the debugger path that the extension will unconditionally launch. This puts the user at a risk, thus this will not be merged like this.

GitMensch:
Thanks for the quick answer. That's an interesting point. Do you think they did not considered that when writing the "Workspace Trust Extension Guide" (or did I miss something there)?

GitMensch:
I've guessed the point is that the user would have to manually click on "run" so there's already a "second step" to code execution, but this is what they say:

> ## What is the guidance for debug extensions? ##
> VS Code will prevent debugging in an untrusted workspace. For this reason, generally, debugging extensions do not need to additionally require trust and should select true for the supported property. However, if your extension provides additional functionality, commands, or settings that are not part of the built-in debugging flow, you should use 'limited' and the above guidance.

The point form above is: debugging is disabled in an untrusted workspace in any case. Isn't that the case any more?
Hm, maybe that's outdated?

WebFreak001:
oh right sorry I didn't check that link - in that case I think it's ok to merge

package.json
```diff
@@ -10,7 +10,7 @@
 		"debug"
 	],
 	"license": "public domain",
-	"version": "0.25.0",
+	"version": "0.25.1",
```
WebFreak001, line 13: needs a changelog entry if you update the version here

WebFreak001 (changes requested):

package.json
GitMensch: ok, I'll split and and a second one with a Changelog entry

GitMensch:
I guess everything is done now? If not: feel free to just commit yourself and I'll learn from inspecting, no need for any credit or anything else.

Issue #278, opened 2021-07-26
Missing quotation for variables with spaces leads to "Could not expand variable: Variable object not found"


martin-fleck-at:
In my program I have a vector variable of a certain type, e.g., `cl::Platform`. When I list all children using GDB with the M2 interface, the name of that variable is reported as `var_1_platforms::_Vector_base<cl::Platform, std::allocator<cl::Platform> >` (note the spaces in the name). But when I try to get the children of that variable (`var-list-children --all-values <name>`) I receive a `Could not expand variable: Variable object not found` exception. I traced the issue to the spaces in the variable name as they break the parsing of the name. This can be easily reproduced on the command line.

Quoting the variable name fixes the issue for me and I noticed that there are already quotation marks around some other commands in the MI2 class, e.g., around expression when we create a variable. I will push a PR adding those quotation marks shortly.

GitMensch:
218d824 should have fixed this, so I assume it can be closed. If there is still an issue please report and we'll reopen.

PR #279, opened 2021-07-26
#278: Add quotation marks around the variable name to support spaces

martin-fleck-at:


WebFreak001:
just quoting them might leave quotation marks in, that will cause the parsing to fail + this behavior could change existing other MI debuggers.

I think it would be best if you can create a function that quotes variables and calls `escape` on the content only if the variables have content that needs to be quoted/escaped (spaces and special characters)

martin-fleck-at:
@WebFreak001 Thank you for your fast reply, I really appreciate it!

I extracted the quoting into a function and use `escape` with surrounding double quotes. Escaping + double quotes is already used in several other places within the class, so I hope that is fine. For now I only do it if we encounter whitespaces which is enough for our use case but I don't know the parsing well enough to know if there are other characters that might break it as well. If you think we should check for more, please let me know.

WebFreak001:
I changed the regex a bit but didn't actually test how it affects the debugging, can you verify if it still works as expected with your issue?

martin-fleck-at:
@WebFreak001 I double checked with our use case and everything still works, fantastic, thank you!

WebFreak001 (approved):

PR #281, opened 2021-07-28
Set as a default debugger for Ada language.

reznikmm:
Hello,

I've tried your extension with Ada programing language and it works well!

Can you enable the extension to be a debugger for Ada? So vscode won't complain that no debugger installed for Ada in code-debug present. And also vscode will found code-debug in the marketplace, I hope.


WebFreak001:
this is the wrong place to add breakpoint languages (refer to #280)

reznikmm:
Breakpoints work fine for Ada already. But I'm trying to fix other issues:

1. If I open an Ada file and click 'Run and Debug" icon (`Ctrl+Shift+D`) VSCode proposes
   > Open a file which can be debugged or run
2. Then vscode also proposes to find a debugger for Ada, but it is unable to find any on the Marketplace.

In [contributes.debuggers](https://code.visualstudio.com/api/references/contribution-points#contributes.debuggers) there is one more property:

> `languages` - those languages for which the debug extension could be considered the "default debugger".

If I assign `["ada"]` to it, then it fixes 1 issue - vscode doesn't propose to "a file which can be debugged or run". I'm not sure if this fixes 2. however.

Here is how to this configured in [MS C/C++ extension](https://github.com/microsoft/vscode-cpptools/blob/c5a7a01889881ab2c096d1c92dc3a7b11f71b718/Extension/package.json#L1445):
```
    "debuggers": [
      {
        "type": "cppdbg",
        "label": "C++ (GDB/LLDB)",
        "languages": [
          "c",
          "cpp",
          "cuda-cpp"
        ],
       ...
```

So it looks like orthogonal to "breakpoints" contribution point.

GitMensch:
I think that's the correct way of handling this. @reznikmm would you please adjust this to include the full list of the breakpoints contribution point?

GitMensch:
@WebFreak001 I'd pull this in the next days if there's no veto from you.

reznikmm:
> I think that's the correct way of handling this. @reznikmm would you please adjust this to include the full list of the breakpoints contribution point?

Sure! I've copied list of all supported language to gdb and lldb-mi debuggers.

reznikmm:
ping @GitMensch

Issue #282, opened 2021-07-29
'threadsRequest' does not send response if 'miDebugger.getThreads()' fails


martin-fleck-at:
The `MI2DebugSession.threadsRequest` does not send a response if `MI2.getThreads` calls fails.

We encountered this problem when using this extension with [Eclipse Theia](https://theia-ide.org/). Specifically, the UI relies on the session to receive a response as otherwise a Promise will never finish, causing subsequent updates to be queued indefinitely.

PR #283, opened 2021-07-29
#282: Ensure we send error response for threads request

martin-fleck-at:


martin-fleck-at:
@WebFreak001 For our own fix we also added a retry behavior for the `threadsRequest` because we noticed that when we start a new debugging session the request sometimes comes a little bit "too early" and we receive an error from GDB:
```
Cannot execute this command while the target is running.
Use the interrupt command to stop the target and try again.
```

Simply re-trying after a short delay (without interrupt) solved that problem for us. So basically we are doing this:
```
protected threadsRequest(response: DebugProtocol.ThreadsResponse, request: DebugProtocol.ThreadsRequest, retry = 5, delay = 100): void {
   ...
   this.miDebugger.getThreads().then(threads => {
      ...
   }).catch(error => {
      if(retry > 1) {
         setTimeout(() => this.threadsRequest(response, request, retry - 1, delay), delay);
      } else {
         this.sendErrorResponse(response, 1, `Could not get threads: ${error}`);
   });
}
```

Do you think it is worth adding that to the code here as well or having a generic "retry" behavior or is that out of scope for the  debug adapter?

WebFreak001 (approved):

WebFreak001:
I think for this PR it's good as it is, for retrying I'm not too sure if it's a good idea just retrying

Issue #284, opened 2021-07-29
Add retry-mechanism to the threadsRequest-method as sometimes it is called too early


martin-fleck-at:
In our use case using GDB as Debugger and Eclipse Theia as frontend, we encountered a scenario where the `threadsRequest` is called too early. In our scenario one debug session is already running and then a second debug session is started. There are 8 threads created in the first session, and 4 in the second session. What happens is that GDB reports that a new thread is created/started and the UI wants to [update the thread information](https://github.com/eclipse-theia/theia/blob/33dea9b48f84eece471357ad271ed2071cc7b012/packages/debug/src/browser/debug-session.tsx#L810) by sending a threads request to the debug adapter,  resulting in the following error from GDB:
```
Cannot execute this command while the target is running.
Use the interrupt command to stop the target and try again.
```

Even without interrupting any running thread, we can query the threads just a few hundred milliseconds later. It is not obvious to me whether this is how it is supposed to work or whether there is any deeper issue underneath. Nevertheless, I am curios whether other people encountered similar cases. We solved our issue by simply re-trying that particular request a few times, so I though that might also be of interest to other people.

PR #285, opened 2021-07-29
#284: Add retry-mechanism to threadsRequest-method for stability

martin-fleck-at:


WebFreak001:
I dislike this fix, retrying after a `Cannot execute this command while the target is running` error seems like bad practice. It should be possible to do a more reliable fix by inspecting the current state of GDB somehow.

I'm in general not a fan of retrying because of bad timing.

GitMensch:
>  It should be possible to do a more reliable fix by inspecting the current state of GDB somehow.

`-thread-info` should be possible to get the state, if it says it is still running, then a temporarily `-interrupt` _could_ be done (but I'd suggest not to do that because it "breaks" `until`/`fin` and friends.
If the state says "still running" then the commands to be executed may be placed in a list and executed as soon as we get the stopped event.

GitMensch:
After some more debugging: I really think it would be good to at least reliable adjust breakpoints while running.
I've tried to enable [GDB's non-stop mode](https://sourceware.org/gdb/onlinedocs/gdb/Non_002dStop-Mode.html#Non_002dStop-Mode) - but it seems the `attach` doesn't even work correctly afterwards.

This may only leave "interrupt", then "continue" (a change I'll not do myself)...

Issue #287, opened 2021-09-05
CWD Tilde handling


vyeevani:
Not sure if this is necessary a feature or just a fix for some so I'll classify it as a feature for now. Fail to find "lldb-mi" when the cwd has a "~" expansion in it.

vyeevani:
According to this, the child process developers rejected handling cwd tilde expansion so it would have to be handled inside this repo: https://github.com/nodejs/node/issues/684

GitMensch:
The feature implementation like that sounds easy:

```js
prog = prog.replace(/^~/, os.homedir());
```

vyeevani:
I’ve made this PR to resolve this with untildify so that the support isn’t hard coded here as a string replacement. https://github.com/WebFreak001/code-debug/pull/288

vyeevani:
@GitMensch do you happen to have write access to get this merged?

GitMensch:
@vyeevani yes, but I won't do it now as @WebFreak001 had more or less the same questions that I had, please see the PR comments.

PR #288, opened 2021-09-06
Added support for tilde to be present on the cwd through untildify

vyeevani:
This PR resolves this issue https://github.com/WebFreak001/code-debug/issues/287

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,26 +42,66 @@
 			}
 		],
 		"breakpoints": [
-			{"language": "c"},
-			{"language": "cpp"},
-			{"language": "d"},
-			{"language": "objective-c"},
-			{"language": "fortran"},
-			{"language": "fortran-modern"},
-			{"language": "fortran90"},
-			{"language": "fortran_free-form"},
-			{"language": "fortran_fixed-form"},
-			{"language": "rust"},
-			{"language": "pascal"},
-			{"language": "objectpascal"},
-			{"language": "ada"},
-			{"language": "nim"},
-			{"language": "arm"},
-			{"language": "asm"},
-			{"language": "vala"},
-			{"language": "crystal"},
-			{"language": "kotlin"},
-			{"language": "zig"}
+			{
+				"language": "c"
+			},
+			{
+				"language": "cpp"
+			},
+			{
+				"language": "d"
+			},
+			{
+				"language": "objective-c"
+			},
+			{
+				"language": "fortran"
+			},
+			{
+				"language": "fortran-modern"
+			},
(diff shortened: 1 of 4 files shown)


vyeevani:
I’ve tested this with tilde in the cwd. Previously I’d see ‘lldb-mi’ not found errors when running the child process spawn for the debugger due to the tilde in the cwd. With this fix I no longer see that errror

package.json
```diff
@@ -42,26 +42,66 @@
...
-			{"language": "objectpascal"},
-			{"language": "ada"},
-			{"language": "nim"},
-			{"language": "arm"},
-			{"language": "asm"},
-			{"language": "vala"},
-			{"language": "crystal"},
-			{"language": "kotlin"},
-			{"language": "zig"}
+			{
```
GitMensch, line 45: Can you please amend your commit with removing that formatting change? The one with the dependency re-order is fine.

vyeevani: Fixed

GitMensch:
[untildify](https://www.npmjs.com/package/untildify) is MIT-licensed as other dependencies (most important the vscode parts) are, so I assume this is fine.
As noted in the issue it would be good to either support tildes in general or not - can you please check if the calls to GDB/LMDB, the executables to debug need that (I guess the easiest option is to create a link to these in your home directory, then setup the paths using the tilde) and if the cwd arguments already internally resolves the tilde already?
If not then we should use untildify also in these places.

Note: untildify currently only works for plain `~` (on all environments including Windows), but the `~user` syntax is currently not supported: https://github.com/sindresorhus/untildify/issues/1 (no show stopper, just something to keep in mind)

vyeevani:
Looks like they don’t need that. At least it seemed to work on my machine but my setup is a bit weird. I’m running this through a dap adapter in emacs. I’ve verified the name in emacs is using a relative path for the executable under debug for lldb it seems to be fine.

src/gdb.ts
```diff
@@ -97,7 +98,7 @@ class GDBDebugSession extends MI2DebugSession {
 				this.sendErrorResponse(response, 102, `Failed to SSH: ${err.toString()}`);
 			});
 		} else {
-			this.miDebugger.load(args.cwd, args.target, args.arguments, args.terminal).then(() => {
+			  this.miDebugger.load(untildify(args.cwd), args.target, args.arguments, args.terminal).then(() => {
```
WebFreak001, line 101: should this really be done in gdb/lldb.ts instead of in the mi engine? You also changed the indentation on most lines by accident.

Issue #294, opened 2021-11-06
Path Substitution Not Properly Escaped


brownts:
Windows/DOS paths in the pathSubstitutions configuration attribute are not escaped before being sent to the debugger.  Thus on Windows, a typical path cannot be used as-is, but instead must be escaped again within the launch.json:

```json
"pathSubstitutions": { "C:\\\\test1": "C:\\\\test2" }
```

Not only is it a nuisance to have to do this (or switch to POSIX style paths), but it is also problematic when attempting to use predefined variables (e.g., ${workspaceFolder}) in the substitution paths.  The predefined VSCode variables use the path format of the host OS, so become unusable in this context.

GitMensch:
It sounds like you could open a PR to fix that, what do you think?

GitMensch:
Fixed with #295 thanks to @brownts.

PR #295, opened 2021-11-09
Correct a number of issues related to path substitutions.

### LINKED ISSUE #293: Path Substitutions Ignored for Attach
The path substitutions are intially handled in "setPathSubstitutions" (i.e., gdb.ts and lldb.ts) by adding specific debugger commands (e.g., "set substitute-path" for GDB), to the list of "extraCommands" in miDebugger.  However, these extra commands are only sent to the backend for the "launch" configuration, but not for the "attach".  It appears the only place the "extraCommands" are referenced are in "MI2.initCommands", which is only invoked from the "MI2.load" API (which corresponds to the "launch" configuration).  It is likely "initCommands" should also be called from the "MI2.connect" and "MI2.attach" APIs as well, which correspond to the "attach" configuration.

brownts:
Fixes #293 and #294.

LLDB path substitutions were not functional since the path commands
were not being sent to the debugger for any of the configurations.
Additionally, the "settings set target.source-map" command was changed
to "settings append target.source-map" to allow for multiple paths to
be specified.  This command also needed to be sent to the debugger as
a "normal command", as it is not part of the MI.

GDB path substitutions were functional for the launch configuration
and SSH, but were not functional for the attach configuration.
Additionally, the paths were not escaped in the "gdb-set
substitute-path" command, therefore Windows/DOS paths were not
preserved correctly.

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -222,15 +222,12 @@ export class MI2 extends EventEmitter implements IBackend {
 			this.process.stderr.on("data", this.stderr.bind(this));
 			this.process.on("exit", (() => { this.emit("quit"); }).bind(this));
 			this.process.on("error", ((err) => { this.emit("launcherror", err); }).bind(this));
-			const commands = [
-				this.sendCommand("gdb-set target-async on", true),
-				this.sendCommand("environment-directory \"" + escape(cwd) + "\"", true)
-			];
+			const promises = this.initCommands(target, cwd, false, true);
 			if (isExtendedRemote) {
-				commands.push(this.sendCommand("target-select " + target));
-				commands.push(this.sendCommand("file-symbol-file \"" + escape(executable) + "\""));
+				promises.push(this.sendCommand("target-select " + target));
+				promises.push(this.sendCommand("file-symbol-file \"" + escape(executable) + "\""));
 			}
-			Promise.all(commands).then(() => {
+			Promise.all(promises).then(() => {
 				this.emit("debug-ready");
 				resolve(undefined);
 			}, reject);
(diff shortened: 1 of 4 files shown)


src/backend/mi2/mi2.ts
```diff
@@ -763,13 +758,13 @@ export class MI2 extends EventEmitter implements IBackend {
 			this.process.stdin.write(raw + "\n");
 	}
 
-	async sendCliCommand(command: string, threadId: number = 0, frameLevel: number = 0) {
+	sendCliCommand(command: string, threadId: number = 0, frameLevel: number = 0): Thenable<MINode> {
```
GitMensch, line 761: Changing this from async to direct call seems like a bigger change, can we move that to a separate PR?

GitMensch (commented):
Other than the `async` change which seems to be unrelated and undocumented the changes look good to me. Maybe an explanation already solves that, too.

Ideally there would be a ChangeLog entry for and package.json bump to 0.26.0.

src/backend/mi2/mi2.ts
brownts: The reason I had to change that was to make the sendCliCommand and sendCommand both have the same type signature.  This is because sendCliCommand is now used in the MI2_LLDB.initCommands to send the extra commands (i.e., "settings append target.source-map" command).  All of the command promises are collected together and thus it seemed to be consistent to combine the commands sent with sendCliCommand together with those sent with sendCommand.

GitMensch: I see. @WebFreak001 Do you see any issues with that or can we pull?
"Kind of related": Could you include a [GH Action to publish the extension](https://github.com/HaaLeo/publish-vscode-extension#readme)? This would allow people with appropriate rights to get an update out of the door without the need for you to join and manually handle that.

WebFreak001: changing from async to returning the Thenable should have no effect other than the return value being returned now, so fine with that.

Publishing automatically sounds like a good idea, debugging the CI if there is a problem will be a pain though

PR #296, opened 2021-11-09
update dependencies and json formats

GitMensch:
@WebFreak001 please check, I found that "too intrusive" to just commit, but the file formats used are deprecated since a while and the components were outdated (especially the ssh2 one).

Note: I'm sadly out of time to test hat version, so "does compile, should work, but not tested"

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -8,23 +8,18 @@
 
 // A task runner that calls a custom npm script that compiles the extension.
 {
-	"version": "0.1.0",
+	"version": "2.0.0",
 
 	// we want to run npm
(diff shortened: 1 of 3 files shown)


WebFreak001:
there was a major bump in ssh2, have you checked the changelog for upgrade notices and checked if everything still works?

GitMensch:
only did a quick test, ssh seems to work as long as I don't use "bootstrap" (I guess I don't use that correctly though...)

WebFreak001 (approved):

GitMensch:
minor follow up for tasks.json with 7d9d147

Issue #277, opened 2021-07-19
Not implemented stop reason (assuming exception): watchpoint-trigger


GitMensch:
Using GDB 10.1, stepping and printing values works in general.

When I add a watch like the following `*(char(*)[8])(0x7fd9166daa8a)` I can directly see its current value and also see the value when GDB halts. But it _seems_ to be more of a monitor as it does not stop when the variable changes. This may also be the reason

When I use the debugging console to manually add it `watch *(char(*)[8])(0x7fd9166daa8a)` then the program does stop as expected and also shows the old and new variant (as expected); but there is a notification and error message, which is bad.

Please catch the event and gracefully stop + update the value in the watch pane (or add it to the watch if it isn't included already).

GitMensch:
Obviously there are [more watchpoint related triggers](https://github.com/bminor/binutils-gdb/blob/master/gdb/mi/mi-common.c):

```c  "breakpoint-hit",
  "watchpoint-trigger",
  "read-watchpoint-trigger",
  "access-watchpoint-trigger",
  "watchpoint-scope",
```

That should be added to the current list in the backend:

https://github.com/WebFreak001/code-debug/blob/c024e2cb0e942db259f8fda7e8ba5bf8e97c195d/src/backend/mi2/mi2.ts#L346-L365

Issue #212, opened 2020-03-04
Error attaching to process


jludwig75:
I get a dialog with the following error trying to attach to a running process:

"Failed to attach: Cannot change this setting while the inferior is running. (from gdb-set target-async on)"

This works just fine with the Microsoft C/C++ extension debugging and with just gdb. The process is running as root on a VM, but Visual Studio Code is also connected and running as root in the VM via VS Code remote ssh. I would really like to get this working, because the Microsoft debugger support doesn't integrate that well with gdb.

This is what I have set in launch.json:
```
        {
            "type": "gdb",
            "request": "attach",
            "name": "Attach to PID",
            "target": "17487",
            "cwd": "${workspaceRoot}",
            "valuesFormatting": "parseText",
            "executable": "/home/vagrant/build/bin/smashfs",
            "showDevDebugOutput": true,
            "printCalls": true
        },
```
This is the output in "DEBUG CONSOLE" when I try to attach:

```1-gdb-set target-async on
2-environment-directory "/home/vagrant/smashfs"
GDB -> App: {"outOfBandRecord":[{"isStream":false,"type":"notify","asyncClass":"thread-group-added","output":[["id","i1"]]}]}
GDB -> App: {"outOfBandRecord":[{"isStream":true,"type":"console","content":"Reading symbols from /home/vagrant/build/bin/smashfs..."}]}
Reading symbols from /home/vagrant/build/bin/smashfs...
GDB -> App: {"outOfBandRecord":[{"isStream":true,"type":"console","content":"done.\n"}]}
done.
GDB -> App: {"outOfBandRecord":[{"isStream":true,"type":"console","content":"Attaching to program: /home/vagrant/build-debug-gcc/bin/smashfs, process 17487\n"}]}
Attaching to program: /home/vagrant/build-debug-gcc/bin/smashfs, process 17487
GDB -> App: {"outOfBandRecord":[{"isStream":false,"type":"notify","asyncClass":"thread-group-started","output":[["id","i1"],["pid","17487"]]}]}
GDB -> App: {"outOfBandRecord":[{"isStream":false,"type":"notify","asyncClass":"thread-created","output":[["id","1"],["group-id","i1"]]}]}
GDB -> App: {"outOfBandRecord":[{"isStream":false,"type":"notify","asyncClass":"thread-created","output":[["id","2"],["group-id","i1"]]}]}
[...]

PR #299, opened 2021-11-11
Add control over continuing on attach.

### LINKED ISSUE #244: gdb attach: make auto-continue configurable: introduce "haltOnAttach"
This is an enhancement request.
In my use-case I am working in Windows, run aarch64 bare-metal software in QEMU started with options "-s -S" -> gdb server, halted at start. I want to source-level debug the remote code using VS code.

When attaching with VS code using Native Debug (and an external aarch64-gdb, remote=true) the code continues immediately after attaching. This is not what I wanted in this case. I wanted that VS code shows me the code location at attach time and does not continue execution automatically.

I found the presumed reason in out/src/gdb.js, GDBDebugSession::attachRequest()
`        this.needContinue = true;`

When changing this line to 
`        this.needContinue = false;`
I get exactly the behaviour I want in this special use-case.

Enhancement request: It would be awesome if there would be a launch.js configuration option "haltOnAttach" or something similar I could use for changing the default behaviour (auto-continue) to "halt".

Great piece of software - thanks!
Jörg

brownts:
Fixes #244.

This uses `stopAtConnect` instead of #244's suggested `haltOnAttach` terminology in order to be consistent with other debug extensions providing this capability.

Allow configuration control over whether execution should continue
after attaching to target.  This is especially useful for remote
debugging when the remote debugger may already be sitting at the
desired location and continuing would not be the appropriate action.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,10 @@ port and optionally hostname in `target`.
 This will attach to the running process managed by gdbserver on localhost:2345. You might
 need to hit the start button in the debug bar at the top first to start the program.
 
+Control over whether the debugger should continue executing on connect can be configured
+by setting `stopAtConnect`.  The default value is `false` so that execution will continue
+after connecting.
+
 ### Using ssh for remote debugging
 
 Debugging using ssh automatically converts all paths between client & server and also optionally
--- a/package.json
+++ b/package.json
@@ -867,6 +867,11 @@
 								"type": "array",
 								"description": "mago commands to run when starting to debug",
 								"default": []
+							},
+							"stopAtConnect": {
+								"type": "boolean",
+								"description": "Whether debugger should stop after connecting to target",
+								"default": false
 							}
 						}
 					}
--- a/src/gdb.ts
+++ b/src/gdb.ts
@@ -30,6 +30,7 @@ export interface AttachRequestArguments extends DebugProtocol.AttachRequestArgum
 	executable: string;
 	remote: boolean;
 	autorun: string[];
+	stopAtConnect: boolean;
 	ssh: SSHArguments;
 	valuesFormatting: ValuesFormattingMode;
 	printCalls: boolean;
@@ -125,7 +126,7 @@ class GDBDebugSession extends MI2DebugSession {
 		this.initDebugger();
 		this.quit = false;
 		this.attached = !args.remote;
-		this.needContinue = true;
+		this.needContinue = !args.stopAtConnect;
 		this.isSSH = false;
 		this.debugReady = false;
 		this.setValuesFormattingMode(args.valuesFormatting);
--- a/src/lldb.ts
+++ b/src/lldb.ts
@@ -28,6 +28,7 @@ export interface AttachRequestArguments extends DebugProtocol.AttachRequestArgum
 	pathSubstitutions: { [index: string]: string };
 	executable: string;
 	autorun: string[];
+	stopAtConnect: boolean;
 	valuesFormatting: ValuesFormattingMode;
 	printCalls: boolean;
 	showDevDebugOutput: boolean;
@@ -112,7 +113,7 @@ class LLDBDebugSession extends MI2DebugSession {
 		this.initDebugger();
 		this.quit = false;
 		this.attached = true;
-		this.needContinue = true;
+		this.needContinue = !args.stopAtConnect;
 		this.isSSH = false;
 		this.debugReady = false;
 		this.setValuesFormattingMode(args.valuesFormatting);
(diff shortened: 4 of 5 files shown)


brownts:
Any reason this can't be merged?

WebFreak001 (approved):

GitMensch:
Just a note from testing: if the program that is attached to currently stays on `getchar` and the setting is `false`, then the extension connects and after some seconds raises exceptions - because the commands used are not possible when the inferior is still running.
When this setting is `true` in this case the debugger shows an exception and the following debug output in the debugger:

~~~
Reading ... 500 shared objects ... from remote target...
Reading /lib64/ld-linux-x86-64.so.2 from remote target...
GDB -> App: {"outOfBandRecord":[{"isStream":true,"type":"console","content":"0x00007fe0f9cea7e0 in __read_nocancel () from target:/usr/lib64/libc.so.6\n"}]}
0x00007fe0f9cea7e0 in __read_nocancel () from target:/usr/lib64/libc.so.6
GDB -> App: {"outOfBandRecord":[{"isStream":false,"type":"exec","asyncClass":"stopped","output":[["frame",[["addr","0x00007fe0f9cea7e0"],["func","__read_nocancel"],["args",[]],["from","target:/usr/lib64/libc.so.6"],["arch","i386:x86-64"]]],["thread-id","1"],["stopped-threads","all"],["core","18"]]}]}
Not implemented stop reason (assuming exception): undefined
14-thread-info
~~~

The program is stopped shortly after `getchar()`, which returns with "0".

brownts:
When `stopAtConnect` is `false`, this is equivalent to the behavior before this option was even added.  Once the editor has seen that the application has stopped, it will start to interrogate it by getting `thread-info`, stack traceback, local variables, etc.  Thus, if you start running shortly after stopping (which is what the default behavior is for an attach), all of these commands initiated by the editor end up getting processed after the application has started running again and you can see failures like this.  That scenario is not unique to the addition of this feature as I saw this happen before this was even added.  I did try and reduce this problem though by the addition of the ContinuedEvent [here](https://github.com/WebFreak001/code-debug/blob/master/src/mibase.ts#L369), but there is still a window of opportunity for this to occur.  I think the only way this can really be minimized is for the extension to maintain the running/stopped state and to send error status back to the editor when it detects this type of condition rather than just blindly forward them on to the debugger.

As for the "Not implemented stop reason", I also see this with GDB on attach when `stopAtConnect` is `true`.  The reason you didn't see it before is because we would continue right away and the issue would be masked.  But for some reason GDB does not provide a stopped reason when you first attach.  And because the extension assumes it was from an exception in this case, you see that shown in VSCode.  It is annoying.  I'm not quite sure if its GDB's fault for not supplying a stop reason in this scenario or the extension for not assuming a better default in this scenario...or maybe both.  Again, this existed prior to this capability having been added (see an example from a couple years ago here: #212), however it may have more visibility now that the option exists not to continue on attach, which will no longer mask this behavior.

GitMensch:
Thanks for the additional explanation here, my thoughts have gone into the same direction and it wasn't a "hey, that feature addition was bad" message (I've also seen similar behavior before) - but to have that written down here is definitely good for later lookup (also by others).

As you wrote that I do wonder if one could reasonable mask the issue if we can't prevent it; but also I'm not sure why this happens at all - if I attach directly via GDB the program stops at the `getchar()`, not "afterwards".

GitMensch:
I've inspected this further - GDB does not return a `reason` in the stop event after `attach`.

Here's the same running from only GDB in mi mode:

~~~
=thread-group-started,id="i1",pid="34521"
=thread-created,id="1",group-id="i1"
=library-loaded,id=... many entries
~"0x00007fa0a626b7e0 in __read_nocancel () from /usr/lib64/libc.so.6\n"
*stopped,frame={addr="0x00007fa0a626b7e0",func="__read_nocancel",args=[],from="/usr/lib64/libc.so.6",arch="i386:x86-64"},thread-id="1",stopped-threads="all",core="1"
~~~

And that just isn't planned for in mi2.ts  https://github.com/WebFreak001/code-debug/blob/e2831d9ab551ea0552ec9ada10a24a9344879fad/src/backend/mi2/mi2.ts#L350-L370 which I assume is likely because that scenario is not mentioned in the [GDB docs](https://sourceware.org/gdb//onlinedocs/gdb/GDB_002fMI-Async-Records.html). I'll send a note to gdb bug mailing list about that and adjust mi2.ts that "no reason =  no problem" (and add the watchpoint triggered events (as simple breakpoints) fixing #277.

brownts:
> I've inspected this further - GDB does not return a `reason` in the stop event after `attach`.

Correct.

> I'll send a note to gdb bug mailing list about that and adjust mi2.ts that "no reason = no problem" (and add the watchpoint triggered events (as simple breakpoints) fixing #277.

Sounds good, thanks!

GitMensch:
Error message is gone, now enjoy `stopAtConnect` without strange errors.

Issue #303, opened 2021-11-23
"start" vs "continue" and the "ui-break-done" event


brownts:
I have a question about the use of the "ui-break-done" event.  This seems to be configured to fire 50ms after any of the launch configurations have completed "loading".  When this fires the "exec-run" command is executed.  However, I noticed that the attach configuration which utilizes the "exec-continue" doesn't behave in this same way.  The attach configurations wait to receive the "configurationDoneRequest" prior to executing the "exec-continue" command.  I'm curious why the launch commands don't behave in the same way, waiting for the configuration to complete prior to executing the "exec-run" command.

It seems by using this arbitrary timeout to delay the "exec-run" command, it may result in not waiting for all of the configuration to have completed prior to starting execution.  There are a number of open issues complaining about not hitting breakpoints set in the UI, and I wonder if this might potentially be the cause...if execution starts prior to the configuration completing.

I was recently experimenting with adding a breakpoint as part of the "configurationDoneRequest", but noticed that execution had started prior to my breakpoint command having been sent, which led me down this path to understand how the launch configurations trigger the execution of the "exec-run" command.

Is there a reason that this delay is used or should the correct behavior be similar to that for the continue which waits until the configuration has completed prior to beginning execution?

WebFreak001:
that's some of the earliest code of this extension, the timeout was inserted probably to make sure, that the UI only calls commands on a valid initialized miDebugger instance. The event is just for the UI to switch into the "debugging is running" view, telling the user that they may now start issuing debug commands, it should not call any commands by itself. The debug host (vscode) may call debug methods on this event though.

50ms is just a nice length for a lot of work to be done, while looking to the user like it is almost instant (in UI design typically operations should not take longer than 100ms to be perceived as instant)

brownts:
Hi @WebFreak001, thanks for the follow-up.

I may be misunderstanding something, but it appears to me that the "ui-break-done" event is only used by the extension.  From my investigation, it is emitted in all of the "launch" scenarios, but there appears to only be a single listener, as shown [here](https://github.com/WebFreak001/code-debug/blob/master/src/backend/mi2/mi2.ts#L384), which is where the "exec-run" command is sent.  That is why it appears to me that the "exec-run" is sent 50ms after "loading" completes, irregardless of whether the UI has finished initializing all of the breakpoints.  In other words, there appears to be a race condition between when the "exec-run" is issued and when the breakpoint configuration completes.

[...]

PR #304, opened 2021-11-25
Update launch to wait for configuration completion before running.

brownts:
This removes the run timer as discussed in #303.

Removed the timer which was previously used to trigger execution of
the run command in the launch configurations and instead updated the
launch configuration to behave like the attach configuration.  This
now waits until the configurationDoneRequest is received before
sending the run command to the debugger.  This is expected to address
race conditions where breakpoints were not being hit due to the
application running prior to the breakpoints having been configured in
the debugger.

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -381,15 +381,13 @@ export class MI2 extends EventEmitter implements IBackend {
 
 	start(): Thenable<boolean> {
 		return new Promise((resolve, reject) => {
-			this.once("ui-break-done", () => {
-				this.log("console", "Running executable");
-				this.sendCommand("exec-run").then((info) => {
-					if (info.resultRecords.resultClass == "running")
-						resolve(undefined);
-					else
-						reject();
-				}, reject);
-			});
+			this.log("console", "Running executable");
+			this.sendCommand("exec-run").then((info) => {
+				if (info.resultRecords.resultClass == "running")
+					resolve(undefined);
+				else
+					reject();
+			}, reject);
 		});
 	}
 
--- a/src/gdb.ts
+++ b/src/gdb.ts
@@ -1,4 +1,4 @@
-import { MI2DebugSession } from './mibase';
+import { MI2DebugSession, RunCommand } from './mibase';
 import { DebugSession, InitializedEvent, TerminatedEvent, StoppedEvent, OutputEvent, Thread, StackFrame, Scope, Source, Handles } from 'vscode-debugadapter';
 import { DebugProtocol } from 'vscode-debugprotocol';
 import { MI2, escape } from "./backend/mi2/mi2";
@@ -56,7 +56,7 @@ class GDBDebugSession extends MI2DebugSession {
 		this.initDebugger();
 		this.quit = false;
 		this.attached = false;
-		this.needContinue = false;
+		this.initialRunCommand = RunCommand.RUN;
 		this.isSSH = false;
 		this.started = false;
 		this.crashed = false;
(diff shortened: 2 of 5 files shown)


tsconfig.json
```diff
@@ -8,6 +8,7 @@
 		],
 		"sourceMap": true,
 		"rootDir": ".",
+		"noFallthroughCasesInSwitch": true,
```
GitMensch, line 11: How is that related to this PR?

WebFreak001: as this PR adds a switch-case I'm not opposed to this if explicit fallthroughs are still allowed (as that's one of the whole points of a switch-case)

GitMensch:
I haven't tested the result but the description and the changes in general (all but the tsconfig) look reasonable to me.

tsconfig.json
GitMensch: I find things like that strange - a switch case should always fall through if there's no explicit `break` (maybe I'm to old already).

brownts: LOL, I didn't realize this was going to be so controversial. :smile: 

> as this PR adds a switch-case I'm not opposed to this if explicit fallthroughs are still allowed (as that's one of the whole points of a switch-case)

Yes, you can suppress this check by placing a @ts-ignore or @ts-expect-error comment on the line before the case statement.  So in the case you really wanted to fall through, that is possible.  I'm not a huge fan of that approach (i.e., using the @ts-ignore or @ts-expect-error) since it's not specific to the fall through check/error, so could mask another errors if they happened to fall on that line as well.

After looking into this a bit more, it seems like this check might be better placed into the lint configuration.  I noticed that there is a [no-switch-case-fall-through](https://palantir.github.io/tslint/rules/no-switch-case-fall-through/) rule available in TSLint, which basically accomplishes the same goal.  This seems more configurable as you can adjust the severity level as well as adding a "falls through" comment to specifically allow the fall through condition, without resorting to potentially masking legitimate errors.

I briefly looked at the lint configuration setup for the native-debug extension as it appeared to have been enabled at some point in the past, but I don't believe it is working any longer.  From my understanding it was previously integrated with `tsc` so that the linting would occur during compilation.  However, it appears the `tsc` plugin (`tslint-language-service` as specified in `tsconfig.json`) is [no longer a  devDependency](https://github.com/WebFreak001/code-debug/commit/271bf42725da3d7a5507998a754117e6dfe8e09d?diff=split#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L965).  Thus, it wasn't even installed on my system.  Furthermore, it seems like this plugin is deprecated and there may be a newer/different approach these days for the integration of linting into the compilation.

I didn't look into the linter configuration much further, but seems it is no longer automatically invoked as part of the compilation.  I ran it from the CLI and noticed I had actually created a few lint violations as part of this change, so I want to clean those up.  However, it may be good to create a new issue to cover the broken lint configuration here and just remove the current "noFallthroughCasesInSwitch" for now.

> I find things like that strange - a switch case should always fall through if there's no explicit break (maybe I'm to old already).

All this does is potentially trigger a warning or error to be omitted (which could also be suppressed) if such a scenario arises.  This allows the developer to confirm their intentions to the compiler/linter that they really want this behavior.  It has been my experience (and I had thought common knowledge) that the "C-style" fall through behavior is almost always unintended.  In fact I'd guess that a legitimate fall through case occurs orders of magnitude less frequently than accidentally forgetting a break statement (and this check exists in many other linters as well), but maybe your experience has been different than mine.

However, as mentioned above, I'll remove this configuration change, fix my linting issues, and then push an update to this PR.  Let me know about the current functionality of the integrated linting and if a new issue should be created for that.

GitMensch: >  Let me know about the current functionality of the integrated linting and if a new issue should be created for that.

Sounds reasonable, at least to discuss and track this.

>  It has been my experience (and I had thought common knowledge) that the "C-style" fall through behavior is almost always unintended. In fact I'd guess that a legitimate fall through case occurs orders of magnitude less frequently than accidentally forgetting a break statement (and this check exists in many other linters as well), but maybe your experience has been different than mine.

I maintain C code where this is intended - and in the places where it is an explicit `/* fall-through */` comment is in. "Yes", I think in most cases that wouldn't be intended and possibly a `break` forgotten, but that's what a `switch` is. If I don't want that than there's the COBOL `EVALUATE` statement ;-)

Thanks for the PR cleanup!

PR #316, opened 2022-03-02
honor debugger args for connect + attach

### LINKED ISSUE #206: 'debugger_args' not honored for 'attach' configuration in gdb

GitMensch:
"should fix #206" - fair warning: untested

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -219,14 +219,15 @@ export class MI2 extends EventEmitter implements IBackend {
 			let args = [];
 			if (executable && !nativePath.isAbsolute(executable))
 				executable = nativePath.join(cwd, executable);
-			if (!executable)
-				executable = "-p";
 			let isExtendedRemote = false;
+			args = this.preargs.concat(this.extraargs || []);
 			if (target.startsWith("extended-remote")) {
 				isExtendedRemote = true;
-				args = this.preargs;
-			} else
-				args = args.concat([executable, target], this.preargs);
+			} else {
+				if (!executable)
+					executable = "-p";
+				args = args.concat([executable, target]);
+			}
 			this.process = ChildProcess.spawn(this.application, args, { cwd: cwd, env: this.procEnv });
 			this.process.stdout.on("data", this.stdout.bind(this));
 			this.process.stderr.on("data", this.stderr.bind(this));
@@ -235,7 +236,8 @@ export class MI2 extends EventEmitter implements IBackend {
 			const promises = this.initCommands(target, cwd, false, true);
 			if (isExtendedRemote) {
 				promises.push(this.sendCommand("target-select " + target));
-				promises.push(this.sendCommand("file-symbol-file \"" + escape(executable) + "\""));
+				if (executable)
+					promises.push(this.sendCommand("file-symbol-file \"" + escape(executable) + "\""));
 			}
 			Promise.all(promises).then(() => {
 				this.emit("debug-ready");
@@ -249,10 +251,9 @@ export class MI2 extends EventEmitter implements IBackend {
 			let args = [];
 			if (executable && !nativePath.isAbsolute(executable))
 				executable = nativePath.join(cwd, executable);
+			args = this.preargs.concat(this.extraargs || []);
 			if (executable)
-				args = args.concat([executable], this.preargs);
-			else
-				args = this.preargs;
+				args = args.concat([executable]);
 			this.process = ChildProcess.spawn(this.application, args, { cwd: cwd, env: this.procEnv });
 			this.process.stdout.on("data", this.stdout.bind(this));
 			this.process.stderr.on("data", this.stderr.bind(this));


WebFreak001:
hm this changes the order of arguments, the preargs are put before the executable now, I'm not sure if gdb arguments are fully unordered.

GitMensch:
~~~
db --help
This is the GNU debugger.  Usage:

    gdb [options] [executable-file [core-file or process-id]]
    gdb [options] --args executable-file [inferior-arguments ...]
~~~

The executable _must_ be the last thing (if you don't use executable + pid, which we currently do via `-p` option; or use `--args`).

WebFreak001 (approved):
though **please test this before merging**.

brownts:
I looked at this a little bit today and I believe the `extraArgs` are still missing from the SSH connections (MI2::ssh) as well as the LLDB attach (MI2_LLDB::attach).

brownts:
I looked at this a little bit today and I believe the `extraArgs` are still missing from the SSH connections (MI2::ssh) as well as the LLDB attach (MI2_LLDB::attach).

brownts:
I was looking at this a bit more today to see how difficult it would be to fix the SSH scenario where the `extraArgs` are not being passed.  I added the `extraArgs` with the `preargs` similar to how this is done in the non-SSH scenarios.  The issue I ran into though is that the parameters are passed differently when running the command locally than they are when running over SSH.  Running the command locally (via `child_process.spawn`) appears to call the process directly.  Also, since the parameters are passed as an array, they appear to be passed directly to the process and there is no need to determine where one parameter ends and the next begins (i.e., spacing).  However, when invoking the command via the `ssh2.exec` command, the command and all of the args are combined (and space separated) as a single string.  This means that the command is parsed on the remote side and thus anything containing spaces, will need to be properly escaped.  However, this behaves differently on Windows than it does on Linux.  For example, I used the following `debugger_args` along with an `autorun` to print the result to verify behavior in the three different scenarios:

- Local (Linux or Windows):
  ```
   debugger_args: ["-iex", "set $foo = \"bar\""],
   autorun: ["print $foo"]
   ```
- SSH to Linux machine:
  ```
   debugger_args: ["-iex", "'set $foo = \"bar\"'"],
   autorun: ["print $foo"]
   ```
- SSH to Windows machine:
  ```
  debugger_args: ["-iex", "\"set $foo = \\\"bar\\\"\""],
  autorun: ["print $foo"]
  ```

The escaping done for the "SSH to Linux machine" scenario didn't work for the "SSH to Windows machine" scenario and vice versa.  Any ideas on how this can be properly escaped for both scenarios over SSH?  It doesn't seem like there should be 3 different versions of `debugger_args` depending on whether you are running this locally or over SSH.  Ideally, the escaping would be done behind the scenes, but I'm not sure there is a universal solution that will work.

GitMensch:
Thank you for investigating again!

> Ideally, the escaping would be done behind the scenes, but I'm not sure there is a universal solution that will work.

Yes, that would be ideal. As there are more users there with the same problem I'd suggest to drop the "how to escape arguments passed to commands" at https://github.com/mscdex/ssh2/issues.

For the "investigation issue part" - if I understand this correctly you've tested the changes and found it working everywhere this issue wants it to "connect + attach" with the exception of LMDB attach (where it should be easy to copy the changes from MI - but I can't test that it works there), correct?
Should this be merged to allow your work be based on it or do you want to tackle it (you're free to copy what you find useful) in which case I'd close the PR?

brownts:
> For the "investigation issue part" - if I understand this correctly you've tested the changes and found it working everywhere this issue wants it to "connect + attach" with the exception of LMDB attach (where it should be easy to copy the changes from MI - but I can't test that it works there), correct?

Hi @GitMensch, I found that it also doesn't work for SSH connections.  I'm not sure if that was considered to be part of this change or not.  I had to add the `extraArgs` to the ssh command to get as far as I had in my previous response.  So, the `extraArgs` are currently not applied on any SSH connection nor on the LLDB attach.

If you want to merge what you have, I can add support for those two remaining conditions in a separate PR (which is probably easier than me trying to apply changes to your current PR).  With my proposed follow-on change, note that the `extraArgs` on the SSH connection may have to be escaped appropriately in order to work unless we can find a solution that would work there as well, but that could be yet another change, if a solution is found that will work.

GitMensch:
Agreed. So looking forward for your SSH change (was not planned to be part) and LMDB fix (was an oversight).

PR #323, opened 2022-03-04
new `ssh.sourceFileMap` configuration for multiple substitutions

### LINKED ISSUE #298: How to apply multiple cwd entries for ssh debugging?
Currently there is only one `cwd` from the launch config mapped to one from `ssh.cwd`.
Commonly multiple ones are needed...

I suggest to keep the current ssh.cwd but mark it as deprecated (this way people's configuration doesn't break after an update but have the change to adjust and not use that when creating a new setup) and add a new configuration `ssh.pathSubstitutions` (map paths between the host machine and the local one; the MS extension names that `sourceFileMap` so we may use that, too - but it should be placed under `ssh` to keep the correct scope).

GitMensch:
fixes #298

also allows to have an actual separate ssh working directory

This is tested and works like a charm (yay, finally I can actually set the working directory - because that's not where the sources are), the only reason for requesting a review (feel free to click on "merge pull request") is that @WebFreak001 previously questioned that setting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 * Breakpoints may be deleted when not recognized correctly #259 fixing #230 (@kvinwang)
 * New `stopAtConnect` configuration #299, #302 (@brownts)
 * New `stopAtEntry` configuration to run debugger to application's entry point #306 (@brownts)
+* New `ssh.sourceFileMap` configuration to allow multiple substitutions between local and ssh-remote and separate ssh working directory #298 (@GitMensch)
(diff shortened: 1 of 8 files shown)


GitMensch:
To be "in line" with the MS setting of the same name

> example: "sourceFileMap": { "/mnt/c": "c:\\" } will map any path returned by the debugger that begins with /mnt/c and convert it to c:\\. You can have multiple mappings in the object but they will be handled in the order provided.

I'll swap it from the current one in this PR which is the other way around. Stay tuned...

GitMensch:
Tested with all different kinds of ssh connections, with/without additional pathSubstitution and also for ssh to win32 - all done -> merged

Issue #329, opened 2022-03-07
Attach to PID does not process "Initial Commands" before attaching


brownts:
The current method used to attach to a process on the local machine is to specify the PID on the command line to the debugger.  However, by doing this, the debugger will attach to the process prior to executing the initial configuration commands (specified in `MI2::initCommands`).  This is inconsistent with remote attach operations (using `target-select remote` or `target-select extended-remote`) which only attach to the remote process after the initial configuration commands have been processed.

The problem with attaching prior to processing the initial configuration commands is that there can be spurious errors that occur because of this.  For example, if a source path mapping is needed to resolve the source code, when the debugger attaches to the process, it will attempt to locate the source file, but won't be able to find it because the source path mappings commands have not yet been processed.  Additionally, this can also lead to errors attempting other commands, such as `gdb-set target-async on` failing with "Cannot change this setting while the inferior is running".

The below output demonstrates both the source path mapping issue as well as the issue with the failing `gdb-set target-async on` command.  I'd suggest that the current behavior be changed to process the initial configuration commands first, and then issue a "target-attach" command afterwards to attach to the local process.

```shell
1-gdb-set target-async on
2-list-features
3-environment-directory "/home/troy/git/hello_world_exe"
4-gdb-set substitute-path "C:\\usr\\src" "/home/troy/git/hello_world_exe/../hello_world_c"
GDB -> App: {"outOfBandRecord":[{"isStream":false,"type":"notify","asyncClass":"thread-group-added","output":[["id","i1"]]}]}
GDB -> App: {"outOfBandRecord":[{"isStream":true,"type":"console","content":"Attaching to process 7801\n"}]}
Attaching to process 7801
GDB -> App: {"outOfBandRecord":[{"isStream":false,"type":"notify","asyncClass":"thread-group-started","output":[["id","i1"],["pid","7801"]]}]}
GDB -> App: {"outOfBandRecord":[{"isStream":false,"type":"notify","asyncClass":"thread-created","output":[["id","1"],["group-id","i1"]]}]}
GDB -> App: {"outOfBandRecord":[{"isStream":true,"type":"console","content":"Reading symbols from /home/troy/git/hello_world_exe/hello_world-gcc..."}]}
Reading symbols from /home/troy/git/hello_world_exe/hello_world-gcc...
GDB -> App: {"outOfBandRecord":[{"isStream":true,"type":"console","content":"done.\n"}]}
done.
[...]

PR #334, opened 2022-03-22
Fix path type detection for differing remote path type.

brownts:
This change will use the CWD to detect the path type where the
debugger is running and use that to determine whether the target was
supplied with an absolute or relative path.

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -5,9 +5,7 @@ import { parseMI, MINode } from '../mi_parse';
 import * as linuxTerm from '../linux/console';
 import * as net from "net";
 import * as fs from "fs";
-import { posix } from "path";
-import * as nativePath from "path";
-const path = posix;
+import * as path from "path";
 import { Client } from "ssh2";
 
 export function escape(str: string) {
@@ -51,8 +49,8 @@ export class MI2 extends EventEmitter implements IBackend {
 	}
 
 	load(cwd: string, target: string, procArgs: string, separateConsole: string): Thenable<any> {
-		if (!nativePath.isAbsolute(target))
-			target = nativePath.join(cwd, target);
+		if (!path.isAbsolute(target))
+			target = path.join(cwd, target);
 		return new Promise((resolve, reject) => {
 			this.isSSH = false;
 			const args = this.preargs.concat(this.extraargs || []);
@@ -166,7 +164,7 @@ export class MI2 extends EventEmitter implements IBackend {
 						this.emit("quit");
 						this.sshConn.end();
 					}).bind(this));
-					const promises = this.initCommands(target, cwd, true, attach);
+					const promises = this.initCommands(target, cwd, attach);
 					promises.push(this.sendCommand("environment-cd \"" + escape(cwd) + "\""));
 					if (procArgs && procArgs.length && !attach)
 						promises.push(this.sendCommand("exec-arguments " + procArgs));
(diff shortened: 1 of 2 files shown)


GitMensch:
Looks like a reasonable approach.
It is based on the following

> Since the CWD is expected to be an absolute path in the debugger's environment

but that assumption is never checked, is it? (I know that I've used a relative cwd and/or `.` in the past, possibly only for the hack "tell the extension to attach, then do the actual attaching via autorun commands", which doesn't need to work forever.

src/backend/mi2/mi2.ts
```diff
@@ -5,9 +5,7 @@ import { parseMI, MINode } from '../mi_parse';
 import * as linuxTerm from '../linux/console';
 import * as net from "net";
 import * as fs from "fs";
-import { posix } from "path";
-import * as nativePath from "path";
-const path = posix;
+import * as path from "path";
```
WebFreak001, line 8: why did you change the definition of path? It's used for SSH path building to always use the posix path.

GitMensch: That was actually a bad hack. Using the explicit `path.posix` in every code below is definitely better.

brownts: Hi @WebFreak001, @GitMensch and I had a discussion on 43cc39e7773f1122ab61cc863256d2264a865e3c where he ran into an issue when the SSH connection was to a Windows machine that the paths were expected to be in Win32 path formatting.  This change was an attempt to "detect" the path type based on the CWD and use the corresponding path manipulation APIs based on that detection.

brownts:
> Looks like a reasonable approach. It is based on the following
> 
> > Since the CWD is expected to be an absolute path in the debugger's environment
> 
> but that assumption is never checked, is it? (I know that I've used a relative cwd and/or `.` in the past, possibly only for the hack "tell the extension to attach, then do the actual attaching via autorun commands", which doesn't need to work forever.

@GitMensch, I was trying to think of a case in which a non-absolute CWD might be supplied and we'd think it was okay to do so.  Your scenario might be the case...although a probably unexpected case.  However, with an attach, we don't use `target` within `initCommands` anyway, so even if it was constructed with the wrong path type, that is probably fine (since it isn't subsequently used).  I just tested the attach with GDB for attaching to a PID as well as a remote attach and they worked fine with a "." CWD, since the subsequent target is not used.

I did notice other places of the code (outside the `initCommands`) that likely don't work with attaching when you don't supply an absolute CWD.   Once place is the LLDB attach (since it issues the `file-exec-and-symbols` command [here](https://github.com/WebFreak001/code-debug/blob/master/src/backend/mi2/mi2lldb.ts#L45)).  Another would likely be the "extended remote".  So your method probably only works in certain attach scenarios but not all, but I don't believe the changes here would have affected those scenarios.  Additionally, I'd think if the changes as described in #329 are implemented, this would probably at least make the GDB attach to PID not work with your method either, if it follows the LLDB approach.

brownts:
If this still doesn't seem sufficient, I could first check that the target path is not a POSIX absolute path and not a Win32 path before attempting to detect the path type from the CWD for the subsequent join.  This might be a little safer in the case where the target is already specified as an absolute path, but the CWD isn't.
```typescript
if (!path.posix.isAbsolute(target) && !path.win32.isAbsolute(target)) {
	const debuggerPath = path.posix.isAbsolute(cwd) ? path.posix : path.win32;
	target = debuggerPath.join(cwd, target);
}
```
I think a Win32 relative path using a leading `/` path separator could be misinterpreted this way as a POSIX absolute path though.  Let me know what you think of this, or if you have a better idea on an approach.

brownts:
Any decision on this?  I'd like to resolve this as HEAD is essentially broken right now for SSH connections due to this problem.  If the fix is to just to do the change as suggested in https://github.com/WebFreak001/code-debug/commit/43cc39e7773f1122ab61cc863256d2264a865e3c, that's fine, but this problem blocking me from addressing some of the other issues.

GitMensch:
_Seems like it did not come through the mail gateway, so here directly again._

I'm undecided.... if @WebFreak001 does not say otherwise then I'll merge tomorrow (= this evening) as I guess you've tested SSH from posix->posix, win32->win32, posix->win32, win32->posix.

If you didn't test some of those please drop a note and I'll test the missing pieces directly after the merge.

GitMensch:
Thanks for working on this.

If you know a test was missing, please drop a note (in general it will be better than my half commit that was there before).

brownts:
I have a setup on Windows now as well.  I did test all 4 combinations that you mention above and they all seem to be working as expected.

PR #338, opened 2022-04-06
Fix extra debugger args for LLDB and SSH.

brownts:
Corrects the LLDB specific attach to also apply the debugger arguments.
Additionally, the SSH connection also now has the extra debugger
arguments applied.


GitMensch:
That's nice. Can you check for necessary ChangeLog notes about the extra args (if I remember there was 1 issue with the released version and multiple PRs). Can be merged directly afterwards!

brownts:
Changelog has now been updated with the corresponding issue and pull requests.

Issue #332, opened 2022-03-08
"go to cursor location" does not work with ssh connection


GitMensch:
The reason is that this call is handled from mi2.ts, while the source conversion is done in mibase.ts.

I have a working version that passes the MI2 instance to MI2DebugSession and then access it to let it do the conversion.
Another alternative would be the other way around - moving the path conversion ssh<->local to be stored and done from the MI2DebugSession.

@WebFreak001 @brownts Are both or one of those "dirty"?
I'm available for doing a PR (I actually have the code for the variants above ;-), just would like to know which way (if one) I shpould take before creating a PR.

GitMensch:
Nope, when I've tried the "go to" feature led to `break-insert -t` (MI2.goto); but the "run to cursor", which I _think_ should use this, just creates a "normal" breakpoint now (see debug output, there's even no need to use the debugger to debug the extension as I did; which showed the call event to be a plain "breakpoint add").

The questions here:

* Why does we get an event of "breakpoint add" instead of goto (or is this just a local issue)?
* Is the change itself "correct" or should it be done different?
* Not sure: Why are some commands handled in MI2 while others are in MI2DebugSession?

GitMensch:
See, we were both confused by this function :-)
But yes, both" run until" and "go to (jump)" are useful; for the later always look out for possible stack trashes or resource leak - best to only jump within the same method and force a return where in an upper method.

The most common use case for jump is going backwards without total `record full` / `record btrace` (or commonly more better: use of `rr`) to inspect a previous state again - or to jump backwards, fix the "code" by assigning variables as they should be, then continue.

I totally agree that the design should be better stated, especially as I'm coming back to the code after weeks without ever "been in completely" I wonder about the same design issues (which partially come from the vscode<->DAP part, partially from the "multiple debuggers" part and partially because "that's the way @WebFreak001 wants it", and partially because there was no better design found yet).
Having more class and method comments _and_ the general design "as good as known" in HACKING.md would help a lot. @brownts Would you mind adding your current view there so we can improve it step by step?

For the issue at hand: if the SSH awareness is in MI2DebugSession then the source file mapping should be done there completely (with the actual code for that in a separate file - as it is now).

GitMensch:
[...]

Issue #330, opened 2022-03-07
Remote vs Extended-Remote Configuration Inconsistent


brownts:
The current way to configure the gdbserver "remote" debugging is to:
- set the `remote` boolean to `true`
- add necessary information in the `target` field (e.g., `1.2.3.4:6789`).
- ```yml
        "request": "attach",
        "executable": "./bin/executable",
        "target": "1.2.3.4:6789",
        "cwd": "${workspaceRoot}",
        "remote": true

The current way to configure the gdbserver "extended remote" debugging is to:
 - set the `remote` boolean to `false` (or not at all)...which seems counter-intuitive
 - add necessary information in the `target` field including `extended-remote` (e.g., `extended-remote 1.2.3.4:6789`)
- ```yml
        "request": "attach",
        "executable": "./bin/executable",
        "target": "extended-remote 1.2.3.4:6789",
        "cwd": "${workspaceRoot}",
        "remote": false

From a user perspective, it seems odd that for `extended-remote` debugging, you would not set the `remote` boolean.  Additionally, having the `target` commands be different (where one requires the `extended-remote`, but the other doesn't require the `remote`) at the beginning is also inconsistent.

It would seem that the `remote` boolean could go away (i.e., deprecate the option) and that the regular remote debugging could just be specified in the same way that the extended remote is, with a `remote` prefix in the `target` property.  I think, at least from a user perspective, this would seem more consistent.

Additionally, internally the paths seem to needlessly diverge since the `extended-remote` is processed via `MI2::attach`, whereas the regular `remote` is processed via `MI2::connect`.  It would seem the logic for both of these could be very similar internally.

Furthermore, standardizing on this might better position the extension to support other capabilities in the future, such as loading a core file (i.e., `target-select core <file>`).

GitMensch:
I totally second that and would favor a PR.

Just one thing to keep in mind: `extended-remote` may also "launch" a new process (you just need the executable and if wanted args _after_ connecting, then use `start` as common. If I remember correctly the current way to do that is to `attach` with `extended-remote server:port` then do the file and args setup along with start manually in autorun commands...

GitMensch:
Related (but outdated): #195 - I'd like to see a clean launch also for `extended-remote`, not sure if we really need more than one function on the debug implementation or if it wouldn't be enough to just pass a boolean / enum for the "types" (normal/remote/extended-remote) and (attach/launch).
[...]

Issue #214, opened 2020-03-07
[Feature Request] Support view register values


hez2010:
I think it's necessary to monitor register values in native debug.  

Currently we have to use `info all-register` to see register values during a debug session, I hope that Native Debug can provide a window to show values in registers.

Thanks.

jscsharp:
In addition to register, it is good to have memory view as well.
I can help to add these feature to it.

nirvdrum:
While not ideal, you can set watch points for registers (e.g., use `$rax` to watch RAX). It would be convenient if these were automatically established.

GitMensch:
see also #242 (so far that's a separate view, but we should provide those as variables)

PR #345, opened 2022-04-22
Allow display of local variables in frames multiple levels deep.

### LINKED ISSUE #342: Local variables not displayed more than 2 stack frames deep
I recently noticed that local variables were not being displayed.  After a little bit of investigation, it appears this only happens if navigating more than 2 stack frames from the top.  This seems to be due to the mapping of stack frames/thread id to the the DAP `variablesReference` (i.e., `level << 16 | threadId`).  This mapping, when navigating more than 2 stack frames deep exceeds the starting point for handle objects (i.e., `VAR_HANDLES_START`), and is incorrectly classified as a variable object in MI2DebugSession::variablesRequest (mibase.ts).

Screenshot_2022-04-19_frame0
Screenshot_2022-04-19_frame1
Screenshot_2022-04-19_frame2

brownts:
Fixes #342.

Remove the arbitrary separation between the variable references used
to represent scopes and those used to represent variable objects.
This implementation uses a common pool of handles to support both
variable object references as well as scope references, thus
eliminating the previous bounded numerical range used for scopes,
which was not large enough to represent stack frames multiple levels
deep, due to thread and level information being encoded in the scope
variable reference value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
-# 0.26.0
+# [Unreleased]
+## Fixed
+* Fixes #342 - Local variables not displayed more than 2 stack frames deep - PR #345 (@brownts)
+
+[Unreleased]: https://github.com/WebFreak001/code-debug/compare/v0.26.0...HEAD
+
+# [0.26.0] - 2022-04-16
 
 * vscode dependency was increased from 1.28 to 1.55 along with the debug-adapter protocol to get rid of some outdated dependencies (@GitMensch)
 * SSH2 module updated from deprecated 0.8.9 to current 1.6.0 (@GitMensch),
@@ -20,6 +26,8 @@
 * allow specification of port/x11port via variable (as numeric string) #265 (@GitMensch)
 * Extra debugger arguments now work in all configurations #316, #338 fixing #206 (@GitMensch, @brownts)
 * Attaching to local PID now performs initialization prior to attaching #341 fixing #329 (@brownts)
+
+[0.26.0]: https://github.com/WebFreak001/code-debug/compare/v0.25.1...v0.26.0
 
 # 0.25.1
 
--- a/src/mibase.ts
+++ b/src/mibase.ts
@@ -16,14 +16,21 @@ class ExtendedVariable {
 	}
 }
 
+class VariableScope {
+	constructor (public readonly name: string, public readonly threadId: number, public readonly level: number) {
+	}
+
+	public static variableName (handle: number, name: string): string {
+		return `var_${handle}_${name}`;
+	}
+}
+
 export enum RunCommand { CONTINUE, RUN, NONE }
 
-const STACK_HANDLES_START = 1000;
-const VAR_HANDLES_START = 512 * 256 + 1000;
-
 export class MI2DebugSession extends DebugSession {
-	protected variableHandles = new Handles<string | VariableObject | ExtendedVariable>(VAR_HANDLES_START);
+	protected variableHandles = new Handles<VariableScope | string | VariableObject | ExtendedVariable>();
 	protected variableHandlesReverse: { [id: string]: number } = {};
+	protected scopeHandlesReverse: { [key: string]: number } = {};
 	protected useVarObjects: boolean;
 	protected quit: boolean;
 	protected attached: boolean;
@@ -176,8 +183,10 @@ export class MI2DebugSession extends DebugSession {
 		try {
 			if (this.useVarObjects) {
 				let name = args.name;
-				if (args.variablesReference >= VAR_HANDLES_START) {
-					const parent = this.variableHandles.get(args.variablesReference) as VariableObject;
+				const parent = this.variableHandles.get(args.variablesReference);
+				if (parent instanceof VariableScope) {
+					name = VariableScope.variableName(args.variablesReference, name);
+				} else if (parent instanceof VariableObject) {
 					name = `${parent.name}.${name}`;
 				}
 
@@ -396,7 +405,23 @@ export class MI2DebugSession extends DebugSession {
 
 	protected scopesRequest(response: DebugProtocol.ScopesResponse, args: DebugProtocol.ScopesArguments): void {
 		const scopes = new Array<Scope>();
-		scopes.push(new Scope("Local", STACK_HANDLES_START + (parseInt(args.frameId as any) || 0), false));
+		const [threadId, level] = this.frameIdToThreadAndLevel(args.frameId);
+
+		const createScope = (scopeName: string, expensive: boolean): Scope => {
+			const key: string = scopeName + ":" + threadId + ":" + level;
+			let handle: number;
+			
+			if (this.scopeHandlesReverse.hasOwnProperty(key)) {
+				handle = this.scopeHandlesReverse[key];
+			} else {
+				handle = this.variableHandles.create(new VariableScope(scopeName, threadId, level));
+				this.scopeHandlesReverse[key] = handle;
+			}
+
+			return new Scope(scopeName, handle, expensive);
+		}
+
+		scopes.push(createScope("Local", false));
 
 		response.body = {
 			scopes: scopes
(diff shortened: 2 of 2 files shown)


GitMensch (commented):
Looks good to me in general. Things I'm not sure about:

1. Shouldn't there be more than the "Local" scope?
2. Is it possible to use the [scope from the DAP](https://microsoft.github.io/debug-adapter-protocol/specification#scope) here instead of an own class?

GitMensch:
See https://github.com/WebFreak001/code-debug/pull/242 for how to acquire variables for the Register scope, which would be a possible follow-up to this change.

brownts:
> 1. Shouldn't there be more than the "Local" scope?

The current implementation only supports the "Local" scope.  Additional scopes could be added, but this PR is to address a bug in the current implementation.  I'd rather keep bug fixes and new functionality separate, from a semantic versioning perspective, as bug fixes can go into patch releases (e.g., 0.26.1), but new functionality should go into a minor release (e.g., 0.27.0).  It seems there are other open issues (e.g., #214) to capture adding an additional scope.

> 2. Is it possible to use the [scope from the DAP](https://microsoft.github.io/debug-adapter-protocol/specification#scope) here instead of an own class?

I assume you are referring to the addition of the `VariableScope` class.  That class is used to map the scope name, debugger thread and stack frame level to the corresponding `variablesReference` value that is supplied in the DAP `Scope` response (see return value of `MI2DebugSession::scopesRequest::createScope` for how this information is used to map to the `variablesReference` supplied in the  DAP `Scope` response).  The previous implementation attempted to encode the debugger's thread and level directly into the `variablesReference` in the DAP `Scope` response, but when stack levels were deep enough the encoded value clashed with the `variablesReference` numerical space used to represent variables.

So, to answer your question, the DAP `Scope` is used (as stated above), but the `VariableScope` is used to hold additional internal information (i.e., thread id, stack frame level) that is not directly represented in the DAP `Scope` response.

GitMensch (approved):

CHANGELOG.md
```diff
@@ -80,3 +84,6 @@ Fixes:
 Added support for pretty printers in variable list (by @gentoo90), enable
 with `"valuesFormatting": "prettyPrinters"` if you have a pretty printer
 to get potentially improved display of variables.
+
+[Unreleased]: https://github.com/WebFreak001/code-debug/compare/v0.26.0...HEAD
+[0.26.0]: https://github.com/WebFreak001/code-debug/compare/v0.25.1...v0.26.0
```
WebFreak001, line 89: I like adding this here, but having the links at the bottom of the file _will_ cause contributors to not update this every time a new release gets added. I would recommend moving this immediately after the version headers.

src/mibase.ts
```diff
@@ -16,14 +16,21 @@ class ExtendedVariable {
 	}
 }
 
-export enum RunCommand { CONTINUE, RUN, NONE }
+class VariableScope {
+	constructor (public readonly name: string, public readonly threadId: number, public readonly level: number) {
+	}
+
+	public static variableName (handle: number, name: string): string {
+		return `var_${handle}_${name}`;
```
WebFreak001, line 24: this changes the variable name that was previously used which was guaranteed a valid C identifier (or at least only contain everything that name could contain) to now also contain colons. Additionally this greatly increases variable name length, especially for nested variables, where I could imagine some debuggers might have some length limitations.

Everything the GDB-MI says about the parameter:
> The name parameter is the string by which the object can be referenced. It must be unique. If ‘-’ is specified, the varobj system will generate a string “varNNNNNN” automatically. It will be unique provided that one does not specify name of that format. The command fails if a duplicate name is found. 

It doesn't specify what characters are valid in there, but I would at least be cautious with other implementations with this. It would probably be better to let GDB decide on the name using `-` and storing that somewhere, though I think that would be better in a separate PR.

Do you think we should replace the : here with something identifier-safe?

WebFreak001 (changes requested):
I like the change, can you check out my two comments?

src/mibase.ts
GitMensch: Why should this be C identifier related? There are other languages supported by GDB which don't follow the same identifier rules (including "exotics" that have hyphenated words). In general GDB can handle every name, as long as you quote it.

If the debugger can be used to get a name (and this also works with other debuggers than GDB) then this would be preferable; otherwise a "plain numeric one" seems to be better.

WebFreak001: hm ok it probably doesn't need to be C identifier related, but the length is definitely increasing a lot with this which I think will push forward a potential issue we had with very deep variables even earlier now.

CHANGELOG.md
brownts: I was looking for more structure in the Changelog and ran across the [Keep a Changelog](https://keepachangelog.com/) and thought it had a lot of good ideas.  I especially like the subsections where you group like changes (fixed, added, deprecated, etc).  I didn't convert the changelog to this format, as I wanted to get your buy-in, but I did add a couple ideas from it.  Notably, I started using the subsections to classify things and did add the "compare" links to the bottom.

If you like this approach, we can add a pre-release check to make sure the links are updated as expected prior to the release (possibly with [this github action](https://github.com/mindsers/changelog-reader-action)), similar to the current checks which make sure all of the release versions are synchronized.  Another option would be to automate part of this (possibly with [this github action](https://github.com/thomaseizinger/keep-a-changelog-new-release)).  I think there are few different ways we could address this concern if we want to keep moving in this direction.

src/mibase.ts
brownts: > this changes the variable name that was previously used which was guaranteed a valid C identifier (or at least only contain everything that name could contain) to now also contain colons. Additionally this greatly increases variable name length, especially for nested variables, where I could imagine some debuggers might have some length limitations.
> 

I don't think I changed this from the previous naming convention, so I'm not sure how this increases the name length.  Are you possibly confusing the parameter `name` here in this static function, with the `name` property in the class?  I just moved the previous functionality (from `variablesRequest`) into a function because I had to use it in `setVariableRequest` as well (for scope-level variables), as that was previously broken too.  In fact, in most cases I was able to make the variable names shorter, since due to using the handles for both the scope and variables, the scope `variablesReference` (which is the "id" encoded into the "var" name) start at a much lower number (i.e., 1000 instead of the previous `level << 16 | threadId`), which once you get a few levels deep turned into a big number.

CHANGELOG.md
WebFreak001: I'm fine with adding it, but the markdown link references at the bottom should really be grouped with the headers. For users viewing the README file without a markdown viewer it'd become much nicer to follow.

These indirect markdown links can be placed anywhere in the file I think, so we can either just put them immediately after the version header or put it at the end of each version block, which kinda makes more sense when reading it as well.

GitMensch:
> I'd rather keep bug fixes and new functionality separate, from a semantic versioning perspective, as bug fixes can go into patch releases (e.g., 0.26.1), but new functionality should go into a minor release (e.g., 0.27.0)

Of course new minor versions can/should also include bug fixes ;-)

I _think_ for 0.26.1 #346 and possibly #332 are the open points, before going on to bigger changes. @brownts I guess you don't want to include #330 soon, do you?
#347 would be useful at any time, @WebFreak001 are you working on this or is it mainly a point on the TODO list?

WebFreak001:
@GitMensch it's just a point on the TODO list, I haven't really touched code-debug code in a while

brownts:
In reference to #347, I do have a unit test github action that I created locally to run the existing unit tests for my own testing, but hadn't pushed it yet.  I can create a PR for that soon.  I haven't done anything towards an integration test though.

GitMensch:
Sounds good! Please do so and let's discuss automated testing in that PR or the referenced issue.

brownts:
Also, to follow-up on your other question, I haven't done much on #330 yet, so it's not ready to go into a near-term release.

GitMensch:
> In reference to #347, I do have a unit test github action that I created locally to run the existing unit tests for my own testing, but hadn't pushed it yet. I can create a PR for that soon. I haven't done anything towards an integration test though.

@brownts: was the unit testing part already pushed with the sourceFileMap changes today or are there more? In any case: can you please PR the unit testing GH action you already have?

PR #353, opened 2022-05-03
apply sourceFileMap for "go to cursor", fixes #332

### LINKED ISSUE #332: "go to cursor location" does not work with ssh connection

GitMensch:
fix #332 by passing MI2 an instance of MI2DebugSession

Note: for reasons I do not know I actually don't reach that code any more!
While the code `response.body.supportsGotoTargetsRequest = true;` is executed and the capability is sent to the debug client - it never gets executed. Instead a breakpoint setting and target continuation is requested and done by code-debug - which is really bad for different reasons including: the breakpoint persists [it is not temporary], the code does a lot more unnecessary stuff and as a side effect: the bug was not reproducible any more because the conversion works fine with the breakpoints already.

I still would like to see a code review - especially as I wasn't sure how to handle it best:

* pass MI2 an instance of MI2DebugSession (as done)
* pass MI2DebugSession to MI2
* pass a callable to do the conversion from MI2DebugSession to MI2 (sounds "cleaner", but I don't know yet how to do it)
* use messaging (I guess that would be bad because we want that do be done directly)

Also, if this PR is merged, a follow-up check should be done if there are more reasonably things to directly call from one class in the other.

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -7,6 +7,7 @@ import * as net from "net";
 import * as fs from "fs";
 import * as path from "path";
 import { Client } from "ssh2";
+import { MI2DebugSession } from "../../mibase";
 
 export function escape(str: string) {
 	return str.replace(/\\/g, "\\\\").replace(/"/g, "\\\"");
@@ -25,7 +26,7 @@ function couldBeOutput(line: string) {
 const trace = false;
 
 export class MI2 extends EventEmitter implements IBackend {
-	constructor(public application: string, public preargs: string[], public extraargs: string[], procEnv: any, public extraCommands: string[] = []) {
+	constructor(private dbg_session: MI2DebugSession, public application: string, public preargs: string[], public extraargs: string[], procEnv: any, public extraCommands: string[] = []) {
 		super();
 
 		if (procEnv) {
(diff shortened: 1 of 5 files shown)


brownts:
Hi @GitMensch,

I'm not a big fan of creating a dependency between MI2 on MI2DebugSession.  MI2DebugSession already has a dependency on MI2 (i.e., the `miDebugger` class attribute).  Thus adding a dependency of MI2 on MI2DebugSession would create a circular dependency.

Additionally, I think there is a simpler solution here.  `MI2::goto` is only called by `MI2DebugSession::gotoTargetsRequest`.  It would seem that MI2DebugSession is the place to perform this conversion prior to invoking `MI2::goto`.  This would be consistent with how the other path conversions are performed as well.  Something like the following:

```diff
	protected gotoTargetsRequest(response: DebugProtocol.GotoTargetsResponse, args: DebugProtocol.GotoTargetsArguments): void {
-		this.miDebugger.goto(args.source.path, args.line).then(done => {
+		const path: string = this.isSSH ? this.sourceFileMap.toRemotePath(args.source.path) : args.source.path;
+		this.miDebugger.goto(path, args.line).then(done => {
			response.body = {

```

GitMensch:
When we only need it there then the move of this function is totally the right thing. Can you check for possibly other functions which have this issue and move them together (+applying the sourceFileMap there)?

It would be some days until I get back to coding typescript.

GitMensch:
No need for those changes - the suggested change was tested and works fine.

PR #358, opened 2022-05-10
add support for envFile

### LINKED ISSUE #258: Add support for `envFile`
Add support for the `envFile` command for loading environment variables from a file [as described in vscode's debugging docs](https://code.visualstudio.com/docs/editor/debugging)

GitMensch:
fix #258

so far untested; it is likely possible to also add unit tests; so work in progress (@ any one, feel free to take over)

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -25,27 +25,30 @@ function couldBeOutput(line: string) {
 const trace = false;
 
 export class MI2 extends EventEmitter implements IBackend {
-	constructor(public application: string, public preargs: string[], public extraargs: string[], procEnv: any, public extraCommands: string[] = []) {
+	constructor(public application: string, public preargs: string[], public extraargs: string[], procEnv: any, envFile: string, public extraCommands: string[] = []) {
 		super();
 
-		if (procEnv) {
-			const env = {};
-			// Duplicate process.env so we don't override it
-			for (const key in process.env)
-				if (process.env.hasOwnProperty(key))
-					env[key] = process.env[key];
+		if (!(envFile || procEnv))
+			return;
 
-			// Overwrite with user specified variables
-			for (const key in procEnv) {
-				if (procEnv.hasOwnProperty(key)) {
-					if (procEnv === undefined)
-						delete env[key];
-					else
-						env[key] = procEnv[key];
-				}
-			}
-			this.procEnv = env;
+		// FIXME: if actuall debugged process is Win32, which we only know _after_ connect,
+		//        then the keys must be read via hashmap using only upper-case for hashing
+		//        ... or entered/compared in upper-case --> possibly postpone this
+		const env = {};
+		// Duplicate process.env so we don't override it
+		for (const key in process.env)
+			if (process.env.hasOwnProperty(key))
+				env[key] = process.env[key];
+
+		// Overwrite with user specified variables
+		if (envFile) {
+			mergeEnv(env, readEnvFile(envFile));
 		}
+		if (procEnv) {
+			mergeEnv(env, procEnv);
+		}
+
+		this.procEnv = env;
 	}
 
 	load(cwd: string, target: string, procArgs: string, separateConsole: string): Thenable<any> {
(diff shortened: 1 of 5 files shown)


src/backend/mi2/mi2.ts
```diff
@@ -25,27 +25,30 @@ function couldBeOutput(line: string) {
...
 
 export class MI2 extends EventEmitter implements IBackend {
-	constructor(public application: string, public preargs: string[], public extraargs: string[], procEnv: any, public extraCommands: string[] = []) {
+	constructor(public application: string, public preargs: string[], public extraargs: string[], procEnv: any, envFile: string, public extraCommands: string[] = []) {
 		super();
 
+		if (!(envFile || procEnv))
+			return;
+
+		// FIXME: if actuall debugged process is Win32, which we only know _after_ connect,
```
brownts, line 34: Do we not really not know what type of process that will be debugged?  The environment support is only applicable to the debugger running on the same machine as the debug adapter (not supported over SSH), so it should be able to determine the system it's running on via `process.platform`.  Not sure I understand this issue.

src/backend/mi2/mi2.ts
```diff
@@ -886,3 +889,32 @@ export class MI2 extends EventEmitter implements IBackend {
 	protected stream;
 	protected sshConn;
 }
+function readEnvFile(filename: string): { [key: string]: string } {
```
brownts, line 892: I would probably recommend taking these functions, along with the functionality in the MI2 constructor above and moving it to its own class (e.g., "Environment").  This would encapsulate this functionality and make it easy to unit test.

src/backend/mi2/mi2.ts
GitMensch: No, we don't know. I'm regularly using gdb to attach to gdbserver on a different machine myself. Works fine as long as the gdb has support for the target's architecture.

brownts: But these environment variables are only being applied to the process where `gdb` is being run, not the process where `gdbserver` is run.  I don't think I understand what the expectation of the environment variables would be in a remote scenario like that.  Are they expected to be visible to the application being debugged?

src/backend/mi2/mi2.ts
GitMensch: Yes, that's a general issue with the existing code - the environment variables are _expected_ to be found in the inferior (this "just happens" when GDB starts it locally). If we cannot make this happen all environment settings should be dropped from all remote scenarios.
I think this should be changed in general to use; and I even forgot if there's a better way than `call setenv (name, value, 1)` [which likely only work for C/C++ code, not everything GDB supports].

brownts: There does appear to be commands in both gdb and lldb for setting environment variables, so using that approach, rather than attempting to set them at the process level might be the way to go.  The environment variable setting commands could just be issued during the MI2::initCommands, similarly to how the path mappings are handled (i.e., extra commands).  The GDB documentation even says these environment variable settings are sent to gdbserver, so should support remote debugging as well (see [here](https://sourceware.org/gdb/onlinedocs/gdb/Environment.html#set-environment)).  This should also work with SSH connections too, since the commands are issued to the debugger itself.

- gdb Add/Remove Environment Variable:
```shell
(gdb) set environment FOO=BAR
(gdb) unset environment FOO
```

- lldb Add/Remove Environment Variable:
```shell
(lldb) settings set target.env-vars FOO=BAR
(lldb) settings remove target.env-vars FOO
```

src/backend/mi2/mi2.ts
GitMensch: >  Are they expected to be visible to the application being debugged?

likely yes, as long as we use [`set environment`](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Environment.html#set-environment) (or its mi equivalent) this will work both for GDB and (at least on GNU/Linux environments) gdbserver (since GDB 8.1); for running processes we`d need to call into `putenv()`, but I think it is fine to say that this is only supported for new inferiors.

GitMensch:
I'm not (possibly anymore) familiar with the test setup - do we have a dev doc how to add tests and run them (outside of CI)?

brownts:
> I'm not (possibly anymore) familiar with the test setup - do we have a dev doc how to add tests and run them (outside of CI)?

The HACKING document does provide a brief overview in the testing section.  The Unit tests use Mocha and I recommended the Mocha Test Explorer extension as an easy way to run the unit tests interactively.

Issue #398, opened 2023-05-23
bad error message if `cwd` points to non-existing directory


GitMensch:
Currently we land in:
https://github.com/WebFreak001/code-debug/blob/26d7d9fe64f222618c3302c323a4da2c426b71a4/src/mibase.ts#L168

I _guess_ there is an option to check `cwd` in both "local" and "remote" case.

WebFreak001:
the simplest option would just be changing the error message to also include a mention about cwd

GitMensch:
Would you mind if I add a check using

```ts
function isDirectoryAvailable(directoryPath: string): Error | null {
  try {
    fs.accessSync(directoryPath, fs.constants.F_OK);
    return null; // Directory exists
  } catch (error) {
    return error; // Return the error object
  }
}
```

to #401 and have a "combined" PR for these issues (this would also go to mibase and be called directly after checking the debugger - as the `ChildProcess.spawn` is used in every case and that takes the cwd.

...hm but the ssh one - did I miss that case when checking for the debugger and should extra handle that for both checks?

PR #401, opened 2023-06-06
check for configured debugger before start to provide a nicer error message

GitMensch:
This is a side issue to #398 (I'd like to provide a similar PR for that).

As I'm currently only "on the browser" I have neither tested that nor run prettier, but I think the current state is enough for a first review.


GitMensch:
That change is easy to test, so those should be added before pulling, too.

GitMensch:
@WebFreak001 Should I rebase - or isn't this likely to get merged?

WebFreak001:
feel free to merge your own commits as well, I think error message improvements and such are always very useful

PR #410, opened 2024-02-07
fix CHANGELOG.md

WebFreak001:


GitMensch:
checked in with 8c926ca1c3d2d4f397cb817cec4d367648e98af9, obviously that was not enough...

WebFreak001:
hm idk what else is broken here

PR #415, opened 2024-03-07
Fixes

oltolm:
Hello,

I tried ND on Windows with VSCode 1.87.1 and GDB 14.2 and ran into a lot of problems. I fixed some of them:

1. the debugger can now display thread names
2. local variables actually work, they were completely broken
3. I changed the default to "prettyPrinters"
4. prevented the error message "Selected thread is running." from displaying, it's not an error
5. cleaned up the code a bit in a separate commit

It still does not display global or member variables, maybe I will fix it later.

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -314,14 +314,14 @@ export class MI2 extends EventEmitter implements IBackend {
 		}
 	}
 
-	onOutputStderr(lines) {
-		lines = <string[]> lines.split('\n');
+	onOutputStderr(str: string) {
+		let lines = str.split('\n');
 		lines.forEach(line => {
 			this.log("stderr", line);
 		});
 	}
 
-	onOutputPartial(line) {
+	onOutputPartial(line: string) {
 		if (couldBeOutput(line)) {
 			this.logNoNewLine("stdout", line);
 			return true;
@@ -329,8 +329,8 @@ export class MI2 extends EventEmitter implements IBackend {
 		return false;
 	}
 
-	onOutput(lines) {
-		lines = <string[]> lines.split('\n');
+	onOutput(str: string) {
+		let lines = str.split('\n');
 		lines.forEach(line => {
 			if (couldBeOutput(line)) {
 				if (!gdbMatch.exec(line))
@@ -391,13 +391,13 @@ export class MI2 extends EventEmitter implements IBackend {
 											case "solib-event":
 											case "syscall-entry":
 											case "syscall-return":
-											// TODO: inform the user
+												// TODO: inform the user
 												this.emit("step-end", parsed);
 												break;
 											case "fork":
 											case "vfork":
 											case "exec":
-											// TODO: inform the user, possibly add second inferior
+												// TODO: inform the user, possibly add second inferior
 												this.emit("step-end", parsed);
 												break;
 											case "signal-received":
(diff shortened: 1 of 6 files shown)


src/mibase.ts
```diff
@@ -283,7 +283,11 @@ export class MI2DebugSession extends DebugSession {
 				response.body.threads.push(new Thread(thread.id, thread.id + ":" + threadName));
 			}
 			this.sendResponse(response);
-		}).catch(error => {
+		}).catch((error: MIError) => {
+			if (error.message === 'Selected thread is running.') {
+				console.error(error);
```
WebFreak001, line 288: `console` stuff is not logged on the client, it's just lost

you probably want to either use the DAP error response or send it to the output using `this.handleMsg("stderr", "...")`

src/backend/mi2/mi2.ts
```diff
@@ -832,10 +830,14 @@ export class MI2 extends EventEmitter implements IBackend {
...
 	}
 
-	async varCreate(expression: string, name: string = "-", frame: string = "@"): Promise<VariableObject> {
+	async varCreate(threadId: number, frameLevel: number, expression: string, name: string = "-", frame: string = "@"): Promise<VariableObject> {
 		if (trace)
 			this.log("stderr", "varCreate");
-		const res = await this.sendCommand(`var-create ${this.quote(name)} ${frame} "${expression}"`);
+		let miCommand = "var-create ";
+		if (threadId != 0) {
+			miCommand += `--thread ${threadId} --frame ${frameLevel}`
```
WebFreak001, line 838:
```suggestion
			miCommand += `--thread ${threadId} --frame ${frameLevel}`;
```

src/backend/mi2/mi2.ts
```diff
@@ -314,23 +314,23 @@ export class MI2 extends EventEmitter implements IBackend {
...
 			this.logNoNewLine("stdout", line);
 			return true;
 		}
 		return false;
 	}
 
-	onOutput(lines) {
-		lines = <string[]> lines.split('\n');
+	onOutput(str: string) {
+		let lines = str.split('\n');
```
WebFreak001, line 333:
```suggestion
		const lines = str.split('\n');
```

src/backend/mi2/mi2.ts
```diff
@@ -314,23 +314,23 @@ export class MI2 extends EventEmitter implements IBackend {
 		}
 	}
 
-	onOutputStderr(lines) {
-		lines = <string[]> lines.split('\n');
+	onOutputStderr(str: string) {
+		let lines = str.split('\n');
```
WebFreak001, line 318:
```suggestion
		const lines = str.split('\n');
```

src/backend/mi2/mi2.ts
oltolm: fixed

src/backend/mi2/mi2.ts
oltolm: fixed

src/backend/mi2/mi2.ts
oltolm: fixed

src/mibase.ts
oltolm: I can see the `console` when debugging the extension. I wanted to ignore this message because it's not an error.

WebFreak001: You will have to either do `sendResponse` or `sendErrorResponse` here though, otherwise it's a hanging response that can cause issues with various DAP servers

oltolm: OK, I used `this.handleMsg` like you suggested.

src/mibase.ts
```diff
@@ -283,7 +283,11 @@ export class MI2DebugSession extends DebugSession {
 				response.body.threads.push(new Thread(thread.id, thread.id + ":" + threadName));
 			}
 			this.sendResponse(response);
-		}).catch(error => {
+		}).catch((error: MIError) => {
+			if (error.message === 'Selected thread is running.') {
+				this.handleMsg("stderr", error.message);
+				return;
```
WebFreak001, line 289: still missing some kind of response, e.g. just `this.sendResponse(response);` if you want it to return as success

oltolm: OK. I also changed it back to `console.error` because I can see it when debugging the extension and it's good enough for me.

WebFreak001 (approved):

oltolm:
There were linting problems, so I ran `npm update --save` and then I had to upgrade typescript `npm install typescript@4.9.3 --save-dev`.

WebFreak001 (changes requested):
can you revert the package lock changes because of `npm update --save` and only run do the typescript update? The other stuff should not change

oltolm:
> can you revert the package lock changes because of `npm update --save` and only run do the typescript update? The other stuff should not change

I reverted the changes and then I just did `npm install typescript@4.9.3 --save-dev`.

WebFreak001 (approved):

PR #419, opened 2024-03-29
set "noImplicitAny" to true and fix the problems

oltolm:
I did some work on code-debug. This is a first PR in a series. I set `noImplicitAny` to true and fixed the problems. I also found a bug in `gdb_expansion.ts` line 226. Not sure if I fixed it correctly.

### DIFF
--- a/src/backend/gdb_expansion.ts
+++ b/src/backend/gdb_expansion.ts
@@ -1,3 +1,4 @@
+import { VariableObject } from "./backend";
 import { MINode } from "./mi_parse";
 
 const resultRegex = /^([a-zA-Z_\-][a-zA-Z0-9_\-]*|\[\d+\])\s*=\s*/;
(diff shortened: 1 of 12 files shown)


src/backend/gdb_expansion.ts
```diff
@@ -223,7 +224,7 @@ export function expandValue(variableCreate: Function, value: string, root: strin
 				val = "Object@" + val;
 			}
 		} else if (typeof val == "string" && val.startsWith("@0x")) {
-			ref = variableCreate(getNamespace("*&" + name.substr));
+			ref = variableCreate(getNamespace("*&" + name.substring(1)));
```
WebFreak001, line 227: ah nice catch, do you know why you remove the first character here? It looks like maybe just passing `name` instead might be correct instead. Sadly there is no test for this to actually verify this.

WebFreak001 (approved):
awesome! Thank you very much for these much needed improvements

GitMensch:
I'm eager to see the rest of the patch series 💪 .

src/backend/gdb_expansion.ts
oltolm: That was just a guess. Feel free to change it.

PR #422, opened 2024-04-06
set strictNullChecks to true in tsconfig.json and fix errors

oltolm:
Another cleanup. I set `strictNullChecks` to `true` in `tsconfig.json` and the fixed all the errors.

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -1,4 +1,4 @@
-import { Breakpoint, IBackend, Thread, Stack, SSHArguments, Variable, RegisterValue, VariableObject, MIError } from "../backend";
+import { Breakpoint, IBackend, Thread, Stack, SSHArguments, Variable, RegisterValue, VariableObject, MIError, ResultBreakpoint } from "../backend";
 import * as ChildProcess from "child_process";
 import { EventEmitter } from "events";
 import { parseMI, MINode } from '../mi_parse';
@@ -48,15 +48,15 @@ export class MI2 extends EventEmitter implements IBackend {
 		}
 	}
 
-	load(cwd: string, target: string, procArgs: string, separateConsole: string, autorun: string[]): Thenable<any> {
+	load(cwd: string, target: string, procArgs: string, separateConsole: string | undefined, autorun: string[]): Thenable<any> {
 		if (!path.isAbsolute(target))
 			target = path.join(cwd, target);
 		return new Promise((resolve, reject) => {
 			this.isSSH = false;
 			const args = this.preargs.concat(this.extraargs || []);
 			this.process = ChildProcess.spawn(this.application, args, { cwd: cwd, env: this.procEnv });
-			this.process.stdout.on("data", this.stdout.bind(this));
-			this.process.stderr.on("data", this.stderr.bind(this));
+			this.process.stdout?.on("data", this.stdout.bind(this));
+			this.process.stderr?.on("data", this.stderr.bind(this));
 			this.process.on("exit", () => this.emit("quit"));
 			this.process.on("error", err => this.emit("launcherror", err));
 			const promises = this.initCommands(target, cwd);
@@ -91,7 +91,7 @@ export class MI2 extends EventEmitter implements IBackend {
 		});
 	}
 
-	ssh(args: SSHArguments, cwd: string, target: string, procArgs: string, separateConsole: string, attach: boolean, autorun: string[]): Thenable<any> {
+	ssh(args: SSHArguments, cwd: string, target: string, procArgs: string, separateConsole: string | undefined, attach: boolean, autorun: string[]): Thenable<any> {
 		return new Promise((resolve, reject) => {
 			this.isSSH = true;
 			this.sshReady = false;
@@ -233,8 +233,8 @@ export class MI2 extends EventEmitter implements IBackend {
 				executable = path.join(cwd, executable);
 			args = this.preargs.concat(this.extraargs || []);
 			this.process = ChildProcess.spawn(this.application, args, { cwd: cwd, env: this.procEnv });
-			this.process.stdout.on("data", this.stdout.bind(this));
-			this.process.stderr.on("data", this.stderr.bind(this));
+			this.process.stdout?.on("data", this.stdout.bind(this));
+			this.process.stderr?.on("data", this.stderr.bind(this));
 			this.process.on("exit", () => this.emit("quit"));
 			this.process.on("error", err => this.emit("launcherror", err));
 			const promises = this.initCommands(target, cwd, true);
@@ -265,8 +265,8 @@ export class MI2 extends EventEmitter implements IBackend {
 			if (executable)
 				args = args.concat([executable]);
 			this.process = ChildProcess.spawn(this.application, args, { cwd: cwd, env: this.procEnv });
-			this.process.stdout.on("data", this.stdout.bind(this));
-			this.process.stderr.on("data", this.stderr.bind(this));
+			this.process.stdout?.on("data", this.stdout.bind(this));
+			this.process.stderr?.on("data", this.stderr.bind(this));
 			this.process.on("exit", () => this.emit("quit"));
 			this.process.on("error", err => this.emit("launcherror", err));
 			const promises = this.initCommands(target, cwd, true);
@@ -450,8 +450,8 @@ export class MI2 extends EventEmitter implements IBackend {
 		return new Promise((resolve, reject) => {
 			this.log("console", "Running executable");
 			this.sendCommand(startCommand).then((info) => {
-				if (info.resultRecords.resultClass == "running")
-					resolve(undefined);
+				if (info.resultRecords?.resultClass == "running")
+					resolve(false);
 				else
 					reject();
 			}, reject);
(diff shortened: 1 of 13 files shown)


src/backend/mi2/mi2.ts
```diff
@@ -589,9 +589,9 @@ export class MI2 extends EventEmitter implements IBackend {
 			let location = "";
 			if (breakpoint.countCondition) {
 				if (breakpoint.countCondition[0] == ">")
-					location += "-i " + numRegex.exec(breakpoint.countCondition.substring(1))[0] + " ";
+					location += "-i " + numRegex.exec(breakpoint.countCondition.substring(1))?.[0] ?? "" + " ";
```
WebFreak001, line 592: we should move the `numRegex.exec(...)` into a separate variable, then check that it matches, then either it must match and otherwise print an error and do nothing, or fall back to some value like `0`

Since we are using `-i`, an argument after it is required and we can't just leave it blank like that. (haven't checked GDB documentation to actually verify this statement though, it just looks like it would work that way)

src/backend/mi2/mi2lldb.ts
```diff
@@ -10,7 +9,7 @@ export class MI2_LLDB extends MI2 {
 		// Since the CWD is expected to be an absolute path in the debugger's environment, we can test
 		// that to determine the path type used by the debugger and use the result of that test to
 		// select the correct API to check whether the target path is an absolute path.
-		const debuggerPath = path.posix.isAbsolute(cwd) ? path.posix : path.win32;
+		const debuggerPath = posix.isAbsolute(cwd) ? posix : win32;
```
WebFreak001, line 12: this looks very weird, I don't think the import separation gains us anything here. In node.js `path` used to be a library that you also always just imported as a whole, although I haven't kept up with the JS world there what is used now.

But considering that we are assigning it to `debuggerPath`, so it can be used like `debuggerPath.join(...)` instead of `path.join(...)` to get the correct `path` for the current platform, I'd much prefer keeping the `path.` prefix

src/frontend/extension.ts
```diff
@@ -1,28 +1,28 @@
...
+import { extname, basename, join } from "path";
+import { tmpdir } from "os";
 
-export function activate(context: vscode.ExtensionContext) {
-	context.subscriptions.push(vscode.workspace.registerTextDocumentContentProvider("debugmemory", new MemoryContentProvider()));
-	context.subscriptions.push(vscode.commands.registerCommand("code-debug.examineMemoryLocation", examineMemory));
-	context.subscriptions.push(vscode.commands.registerCommand("code-debug.getFileNameNoExt", () => {
-		if (!vscode.window.activeTextEditor || !vscode.window.activeTextEditor.document || !vscode.window.activeTextEditor.document.fileName) {
-			vscode.window.showErrorMessage("No editor with valid file name active");
+export function activate(context: ExtensionContext) {
```
WebFreak001, line 7: keep the `vscode` import, it's the recommended style from the vscode docs and also makes finding symbols much easier than global imports. In fact I think all of these imports that you changed are much better in a namespace and shouldn't be unwrapped like this. Especially generic stuff like `join` or `connect` doesn't make much sense to import on its own and just makes the caller code less readable.

src/gdb.ts
```diff
@@ -68,8 +69,8 @@ class GDBDebugSession extends MI2DebugSession {
 		this.started = false;
 		this.crashed = false;
 		this.setValuesFormattingMode(args.valuesFormatting);
-		this.miDebugger.printCalls = !!args.printCalls;
-		this.miDebugger.debugOutput = !!args.showDevDebugOutput;
+		this.miDebugger.printCalls = args.printCalls;
```
WebFreak001, line 72: `args` is user supplied and may be literally any JSON type. (even though it will show type errors to the user in vscode) Keep the boolean cast to make sure we don't break any type constraints in typescript from here.

src/gdb.ts
```diff
@@ -112,8 +113,8 @@ class GDBDebugSession extends MI2DebugSession {
 		this.initialRunCommand = args.stopAtConnect ? RunCommand.NONE : RunCommand.CONTINUE;
 		this.isSSH = false;
 		this.setValuesFormattingMode(args.valuesFormatting);
-		this.miDebugger.printCalls = !!args.printCalls;
-		this.miDebugger.debugOutput = !!args.showDevDebugOutput;
+		this.miDebugger.printCalls = args.printCalls;
+		this.miDebugger.debugOutput = args.showDevDebugOutput;
```
WebFreak001, line 117: ditto

src/lldb.ts
```diff
@@ -63,8 +64,8 @@ class LLDBDebugSession extends MI2DebugSession {
 		this.started = false;
 		this.crashed = false;
 		this.setValuesFormattingMode(args.valuesFormatting);
-		this.miDebugger.printCalls = !!args.printCalls;
-		this.miDebugger.debugOutput = !!args.showDevDebugOutput;
+		this.miDebugger.printCalls = args.printCalls;
+		this.miDebugger.debugOutput = args.showDevDebugOutput;
```
WebFreak001, line 68: ditto

src/lldb.ts
```diff
@@ -107,8 +108,8 @@ class LLDBDebugSession extends MI2DebugSession {
 		this.initialRunCommand = args.stopAtConnect ? RunCommand.NONE : RunCommand.CONTINUE;
 		this.isSSH = false;
 		this.setValuesFormattingMode(args.valuesFormatting);
-		this.miDebugger.printCalls = !!args.printCalls;
-		this.miDebugger.debugOutput = !!args.showDevDebugOutput;
+		this.miDebugger.printCalls = args.printCalls;
+		this.miDebugger.debugOutput = args.showDevDebugOutput;
```
WebFreak001, line 112: ditto

src/mago.ts
```diff
@@ -64,8 +65,8 @@ class MagoDebugSession extends MI2DebugSession {
 		this.started = false;
 		this.crashed = false;
 		this.setValuesFormattingMode(args.valuesFormatting);
-		this.miDebugger.printCalls = !!args.printCalls;
-		this.miDebugger.debugOutput = !!args.showDevDebugOutput;
+		this.miDebugger.printCalls = args.printCalls;
+		this.miDebugger.debugOutput = args.showDevDebugOutput;
```
WebFreak001, line 69: ditto

src/mago.ts
```diff
@@ -86,8 +87,8 @@ class MagoDebugSession extends MI2DebugSession {
 		this.initialRunCommand = args.stopAtConnect ? RunCommand.NONE : RunCommand.CONTINUE;
 		this.isSSH = false;
 		this.setValuesFormattingMode(args.valuesFormatting);
-		this.miDebugger.printCalls = !!args.printCalls;
-		this.miDebugger.debugOutput = !!args.showDevDebugOutput;
+		this.miDebugger.printCalls = args.printCalls;
+		this.miDebugger.debugOutput = args.showDevDebugOutput;
```
WebFreak001, line 91: ditto

src/mibase.ts
```diff
@@ -84,9 +84,9 @@ export class MI2DebugSession extends DebugSession {
 				if (process.platform != "win32")
 					this.handleMsg("stderr", "Code-Debug WARNING: Utility Command Server: Error in command socket " + err.toString() + "\nCode-Debug WARNING: The examine memory location command won't work");
 			});
-			if (!fs.existsSync(systemPath.join(os.tmpdir(), "code-debug-sockets")))
-				fs.mkdirSync(systemPath.join(os.tmpdir(), "code-debug-sockets"));
-			this.commandServer.listen(this.serverPath = systemPath.join(os.tmpdir(), "code-debug-sockets", ("Debug-Instance-" + Math.floor(Math.random() * 36 * 36 * 36 * 36).toString(36)).toLowerCase()));
+			if (!existsSync(join(tmpdir(), "code-debug-sockets")))
```
WebFreak001, line 87: keep `systemPath`, it's written explicitly to differentiate that this is `path` using separators using the host OS config, rather than the debuggerPath, which uses separates on the platform that the debugger is running on

src/mibase.ts
```diff
@@ -188,7 +188,7 @@ export class MI2DebugSession extends DebugSession {
 		else
 			this.miDebugger.stop();
 		this.commandServer.close();
-		this.commandServer = undefined;
+		// this.commandServer = undefined;
```
WebFreak001, line 191: remove line

src/mibase.ts
```diff
@@ -706,6 +706,15 @@ export class MI2DebugSession extends DebugSession {
 	}
 
 	protected override evaluateRequest(response: DebugProtocol.EvaluateResponse, args: DebugProtocol.EvaluateArguments): void {
+		// TODO: evaluate the expression in the global scope
+		if (!args.frameId) {
+			response.body = {
+				variablesReference: 0,
+				result: ""
+			};
+			this.sendResponse(response);
```
WebFreak001, line 715: sendErrorResponse instead of plain sendResponse

WebFreak001 (changes requested):

src/backend/mi2/mi2.ts
oltolm: done

src/backend/mi2/mi2lldb.ts
oltolm: I reverted all changes to imports.

src/mibase.ts
oltolm: done

src/mibase.ts
oltolm: done

oltolm:
I fixed everything.

oltolm:
Is there still something to do to get it accepted?

PR #426, opened 2024-05-27
add support to Log Message breakpoint

HenryRiley0:
add log message breakpoint function.

image

image

image

image

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -15,6 +15,13 @@ export function escape(str: string) {
 const nonOutput = /^(?:\d*|undefined)[\*\+\=]|[\~\@\&\^]/;
 const gdbMatch = /(?:\d*|undefined)\(gdb\)/;
 const numRegex = /\d+/;
+const logMsgMatch = /(^\$[0-9]*[\ ]*=[\ ]*)(.*)/;
+const logReplaceTest = /{([^}]*)}/g;
+
+function isPositiveInteger(str: string): boolean {
+	const regex = /^[1-9]\d*$/;
+	return regex.test(str);
+}
 
 function couldBeOutput(line: string) {
 	if (nonOutput.exec(line))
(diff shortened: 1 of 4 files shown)


src/backend/mi2/mi2.ts
```diff
@@ -576,6 +619,26 @@ export class MI2 extends EventEmitter implements IBackend {
...
+		let commands = "";
+
+		while ((m = regex.exec(command))) {
+			if (m.index === regex.lastIndex) {
+				regex.lastIndex++;
+			}
+
+			m.forEach((match, groupIndex) => {
+				if(groupIndex === 1){
+					commands += "\"print " + match + "\" ";
```
WebFreak001, line 634: probably want to just access `m[1]` instead of iterating and checking for an exact index?

might want to escape `match` as command argument as well?

src/backend/mi2/mi2.ts
```diff
@@ -591,12 +654,17 @@ export class MI2 extends EventEmitter implements IBackend {
...
+						resolve([false, undefined]);
+						return;
+					}
 					const match = numRegex.exec(breakpoint.countCondition)[0];
 					if (match.length != breakpoint.countCondition.length) {
 						this.log("stderr", "Unsupported break count expression: '" + breakpoint.countCondition + "'. Only supports 'X' for breaking once after X times or '>X' for ignoring the first X breaks");
 						location += "-t ";
 					} else if (parseInt(match) != 0)
-						location += "-t -i " + parseInt(match) + " ";
+						location += "-t -i " + (parseInt(match) - 1) + " ";
```
WebFreak001, line 667: change in behavior, revert or put into separate PR with changelog if there is a reason for this

src/backend/mi2/mi2.ts
```diff
@@ -373,6 +405,17 @@ export class MI2 extends EventEmitter implements IBackend {
...
 											case "breakpoint-hit":
 												this.emit("breakpoint", parsed);
+												this.logMsg.forEach((brk, index)=>{
+													if(parsed.outOfBandRecord[0].output[0][1] == "breakpoint-hit" && parsed.outOfBandRecord[0].output[2][1] == brk.id){
+														this.logMsgNum = brk.id;
+														this.logMsgVar = brk?.logMessage;
+														const matches = this.logMsgVar.match(logReplaceTest);
+														const count = matches ? matches.length : 0;
+														this.logMsgRplNum = count;
+														this.logMsgVarProcess = this.logMsgVar.replace(logReplaceTest, "placeHolderForVariable");
```
WebFreak001, line 415: this seems rather random and temporary.

Can you move the functionality here + the new variables into a helper class, so it's easier to reason about the logic?

WebFreak001 (changes requested):

PR #427, opened 2024-05-29
add support to Log Message breakpoint

HenryRiley0:
#426  @WebFreak001, thanks for your advice, i modified the code, makes LogMessage var and function in a class.


WebFreak001:
cool! can you add a sample, maybe a sample screenshot/gif in the README or other docs file to show off this feature?

src/backend/mi2/mi2.ts
```diff
@@ -24,6 +24,50 @@ function couldBeOutput(line: string) {
...
+	}
+
+	logMsgProcess(parsed){
+		this.logMsgBrkList.forEach((brk)=>{
+			if(parsed.outOfBandRecord[0].output[0][1] == "breakpoint-hit" && parsed.outOfBandRecord[0].output[2][1] == brk.id){
+				this.logMsgVar = brk?.logMessage;
+				const matches = this.logMsgVar.match(this.logReplaceTest);
+				const count = matches ? matches.length : 0;
+				this.logMsgRplNum = count;
+				this.logMsgVarProcess = this.logMsgVar.replace(this.logReplaceTest, "placeHolderForVariable");
```
WebFreak001, line 64: I'm not a huge fan of the magic string here and would much prefer if we could instead maybe split by the regex, so it's not possible to have this value accidentally in the output message (even if very unlikely)

WebFreak001:
btw you can just push to the `new` branch that you made and the PR will automatically update on github, you don't have to create a new one when you change things

HenryRiley0:
> btw you can just push to the `new` branch that you made and the PR will automatically update on github, you don't have to create a new one when you change things

thank you. i fix the lint error, add a gif in readme

WebFreak001 (approved):

PR #428, opened 2024-05-31
Bugfix

HenryRiley0:
1.fix  implicitly type error of LogMessage
2.fix gdb check error, checkCommand success should not warning "Configured debugger ${dbgCommand} not found"  

```
	protected checkCommand(debuggerName: string): boolean {
		try {
			const command = process.platform === 'win32' ? 'where' : 'command -v';
			execSync(`${command} ${debuggerName}`, { stdio: 'ignore' });
			return true;
		} catch (error) {
			return false;
		}
	}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ Versioning].
 
 ### Added
 
+- fix gdb check error when debug beginning ([@henryriley0])
+- fix implicitly type error in log message when build vsix ([@henryriley0])
 - check for configured debugger before start to provide a nicer error message
   ([@GitMensch])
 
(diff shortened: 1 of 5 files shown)


HenryRiley0:
the newest code will fail, please consider merge this pr

GitMensch:
I'd like to pull that in, can you please add your notes to the Changelog as well?

HenryRiley0:
> I'd like to pull that in, can you please add your notes to the Changelog as well?

ok, i will put a changelog later

HenryRiley0:
> I'd like to pull that in, can you please add your notes to the Changelog as well?

change log added

PR #435, opened 2024-08-07
fix static variable not showing

HenryRiley0:
the static member of Class People does not show 
Log

```
GDB -> App: {"token":17,"outOfBandRecord":[],"resultRecords":{"resultClass":"done","results":[["value","{stu = {name = 0x0, age = 0, score = 0}, pt = {name = 0x0, learntime = 0}, static borth = 10}"]]}}
-gdb-exit
```
image

does not show:

fixed:
fixed

### DIFF
--- a/src/backend/gdb_expansion.ts
+++ b/src/backend/gdb_expansion.ts
@@ -197,6 +197,7 @@ export function expandValue(variableCreate: (arg: VariableObject | string, optio
 
 	parseResult = (pushToStack: boolean = false) => {
 		value = value.trim();
+		value = value.replace(/^static /, "");
 		const variableMatch = resultRegex.exec(value);
 		if (!variableMatch)
 			return undefined;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Versioning].
 
 ### Added
 
+- add static variable support ([@henryriley0])
 - fix gdb check error when debug beginning ([@henryriley0])
 - fix implicitly type error in log message when build vsix ([@henryriley0])
 - check for configured debugger before start to provide a nicer error message


HenryRiley0:
i tested 'register、const、mutable、volatile、static', only static variable will return a modifier.  it seems only static variable will has a modifier，for emphasize the variable's life time is  different from loacal vairable. like below:

image
> 18-stack-list-variables --thread 1 --frame 0 --simple-values
GDB -> App: {"token":18,"outOfBandRecord":[],"resultRecords":{"resultClass":"done","results":[["variables",[[["name","regVar"],["type","int"],["value","10"]],[["name","p1"],["type","People"]],[["name","constVar"],["type","const int"],["value","20"]],[["name","volatileVar"],["type","volatile int"],["value","30"]],[["name","testObj"],["type","Test"]]]]]}}
19-data-evaluate-expression "p1"
GDB -> App: {"token":19,"outOfBandRecord":[],"resultRecords":{"resultClass":"done","results":[["value","{static borth = 10}"]]}}
20-data-evaluate-expression "testObj"
GDB -> App: {"token":20,"outOfBandRecord":[],"resultRecords":{"resultClass":"done","results":[["value","{mutableVar = 1}"]]}}

static variable and 'anonymous union' variable name has a space.  if dont have a space "static borth = 10" and "anonymous union" will test fail.
```
	parseResult = (pushToStack: boolean = false) => {
		value = value.trim();
		const variableMatch = resultRegex.exec(value);
		if (!variableMatch)
			return undefined;
		value = value.substring(variableMatch[0].length).trim();
```

GitMensch:
what about something like:

```ts
const resultRegex = /\s*(static )?([a-zA-Z_\-][a-zA-Z0-9_\-]*|\[\d+\])\s*=\s*/;

	parseResult = (pushToStack: boolean = false) => {
		const variableMatch = resultRegex.exec(value);
		if (!variableMatch)
			return undefined;
		value = variableMatch[1];
```

HenryRiley0:
> what about something like:
> 
> ```ts
> const resultRegex = /\s*(static )?([a-zA-Z_\-][a-zA-Z0-9_\-]*|\[\d+\])\s*=\s*/;
> 
> 	parseResult = (pushToStack: boolean = false) => {
> 		const variableMatch = resultRegex.exec(value);
> 		if (!variableMatch)
> 			return undefined;
> 		value = variableMatch[1];
> ```
> 
> although from the output presented, we can drop the leading `\s*` and replace the others by a single space.

thanks for you adivce, modified resultRegex, works fine.
image
image

src/backend/gdb_expansion.ts
```diff
@@ -229,7 +229,13 @@ export function expandValue(variableCreate: (arg: VariableObject | string, optio
 		} else if (typeof val == "string" && val.startsWith("<...>")) {
 			ref = variableCreate(getNamespace(name));
 			val = "...";
+		} else {
+			let evaluateName = getNamespace(name);
+			if(typeof evaluateName == "string" && evaluateName.includes("static ")) {
+				evaluateName = evaluateName.replace("static ", "");
+			}
```
GitMensch, line 236: You create a local variable here and adjust but never use it - or what did I miss?

src/mibase.ts
```diff
@@ -432,7 +432,11 @@ export class MI2DebugSession extends DebugSession {
 
 	protected override async variablesRequest(response: DebugProtocol.VariablesResponse, args: DebugProtocol.VariablesArguments): Promise<void> {
 		const variables: DebugProtocol.Variable[] = [];
-		const id: VariableScope | string | VariableObject | ExtendedVariable = this.variableHandles.get(args.variablesReference);
+		let id: VariableScope | string | VariableObject | ExtendedVariable = this.variableHandles.get(args.variablesReference);
+
+		if(typeof id == "string" && id.includes("static ")) {
+			id = id.replace("static ", "");
+		}
```
GitMensch, line 439: please move that to the place where that `id` with type string is actually used in this function, this both the review and following the code later an easier task

GitMensch (changes requested):
Thanks for the update! some things are still unclear to me, please go on.

src/backend/gdb_expansion.ts
HenryRiley0: Thanks for the reminder, this piece of code is redundant.

src/mibase.ts
HenryRiley0: thanks, i moved this handle to  'typeof id == "string" ' branch

GitMensch: Ah, I see. I suggest to just use `const string name = id.replace(/^static /, "");` (replace also checks for the include) + using name instead of id in the two places below - this way you can also leave the "id" definition as is.
I'll pull that afterwards, so we can go to the next PR.

Thanks for your work on this!

src/mibase.ts
```diff
@@ -531,6 +531,7 @@ export class MI2DebugSession extends DebugSession {
 				this.sendErrorResponse(response, 1, `Could not expand variable: ${err}`);
 			}
 		} else if (typeof id == "string") {
+			id.replace(/^static /, "");
```
GitMensch, line 534: I wasn't aware that this changes the content of the `id` object and that this can be done to a const

HenryRiley0: > I wasn't aware that this changes the content of the `id` object and that this can be done to a const

The code is useless; const variables do not have a const modifier.

GitMensch:
Nice simple one in the end! We should add some automated testing for that, but I am currently not sure how that works -so I've merged it.
If you can add tests, please do so.

HenryRiley0:
> Nice simple one in the end! We should add some automated testing for that, but I am currently not sure how that works -so I've merged it. If you can add tests, please do so.

ok, i will add static tests

PR #434, opened 2024-08-05
mi2: Follow tsc type recommendations for some parameters

dd86k:
Since tsc was emitting errors, I decided to apply the suggestions to mi2.ts to (try to) help understand the intent of the code and make the PR. (I know, lazy PR, and didn't check for other PRs).

Without this patch, I cannot launch the addon for debugging.

Open to comments/suggestions/edits/rejection.

EDIT: Added reason, removed some cruft.

PR #436, opened 2024-08-07
Support GDB frame filters

JacquesLucke:
### Overview

This adds a new `frameFilters` option for GDB. When enabled, `enable-frame-filters` is passed to GDB which will run [frame filters](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Frame-Filter-API.html) before they are passed to the extension. 

By default, the option is off to avoid regressions in existing setups. If no issues are found over time, it may make sense to enable them by default. If a user has custom frame filters, it probably makes sense that they should also show up in vscode. It's still always possible to access the full backtrace in the `Debug Console`.

### Motivation

I want to use these filters to simplify debugging when working on Blender. They can greatly simplify the backtraces making them easier to scan in the majority of use-cases. See the example below which shows the original and simplified backtrace. Note that this is still a fairly simple case and the backtrace may become way deeper in some situations.

Before:
image
After:
image

### Details

For some reason, GDB does not provide the `@frame.fullname` property for stack frames when frame filters are used. However, in my case `@frame.file` was provided and contained the absolute path too. So this is now used as a fallback.

Previously, the stack frame name was always `function_name@address`. However, with frame filters, the address might not always be known. I changed it so that the `@address` part is omitted when the address is unknown.

### DIFF
--- a/src/gdb.ts
+++ b/src/gdb.ts
@@ -17,6 +17,7 @@ export interface LaunchRequestArguments extends DebugProtocol.LaunchRequestArgum
 	stopAtEntry: boolean | string;
 	ssh: SSHArguments;
 	valuesFormatting: ValuesFormattingMode;
+	frameFilters: boolean;
 	printCalls: boolean;
 	showDevDebugOutput: boolean;
 }
@@ -35,6 +36,7 @@ export interface AttachRequestArguments extends DebugProtocol.AttachRequestArgum
 	stopAtEntry: boolean | string;
 	ssh: SSHArguments;
 	valuesFormatting: ValuesFormattingMode;
+	frameFilters: boolean;
 	printCalls: boolean;
 	showDevDebugOutput: boolean;
 }
@@ -69,6 +71,7 @@ class GDBDebugSession extends MI2DebugSession {
 		this.started = false;
 		this.crashed = false;
 		this.setValuesFormattingMode(args.valuesFormatting);
+		this.miDebugger.frameFilters = !!args.frameFilters;
 		this.miDebugger.printCalls = !!args.printCalls;
 		this.miDebugger.debugOutput = !!args.showDevDebugOutput;
 		this.stopAtEntry = args.stopAtEntry;
@@ -113,6 +116,7 @@ class GDBDebugSession extends MI2DebugSession {
 		this.initialRunCommand = args.stopAtConnect ? RunCommand.NONE : RunCommand.CONTINUE;
 		this.isSSH = false;
 		this.setValuesFormattingMode(args.valuesFormatting);
+		this.miDebugger.frameFilters = !!args.frameFilters;
 		this.miDebugger.printCalls = !!args.printCalls;
 		this.miDebugger.debugOutput = !!args.showDevDebugOutput;
 		this.stopAtEntry = args.stopAtEntry;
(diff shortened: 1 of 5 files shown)


src/gdb.ts
```diff
@@ -55,7 +57,7 @@ class GDBDebugSession extends MI2DebugSession {
 
 	protected override launchRequest(response: DebugProtocol.LaunchResponse, args: LaunchRequestArguments): void {
 		const dbgCommand = args.gdbpath || "gdb";
-		if (this.checkCommand(dbgCommand)) {
+		if (!this.checkCommand(dbgCommand)) {
```
WebFreak001, line 60: how was this not broken before?

WebFreak001 (changes requested):
thanks for the PR! Just a question about that one kinda very confusing fundamental change

src/gdb.ts
JacquesLucke: Yeah, I was quite confused by this too. The check seems to be broken since its introduction in d559072646d27418211a7e1d0605940ef4ec7581. It was not released since then, so I guess it just was not actually tested?

See how `checkCommand` returns true in success and false on failure.

WebFreak001: cc @GitMensch 

I'll try this out locally

HenryRiley0: i put a pr about this,  #428  https://github.com/WebFreak001/code-debug/pull/428/commits/2869f942a461410004199b4018849f0505766e2f
this is a bug after 0.27.0 in master branch, https://github.com/WebFreak001/code-debug/commit/d559072646d27418211a7e1d0605940ef4ec7581

GitMensch: solved, we really should try having some more automated checks that we can run on CI...

GitMensch:
@JacquesLucke can you please rebase and add your name entry to the Changelog?

JacquesLucke:
I updated the patch.

GitMensch:
@WebFreak001 @JacquesLucke Do you think that we really need to disable frame-filters per default?
GDB will only _load_ frame-filters if it was requested (either on commands setup in launch.json, on debug-console, or from auto-loaded gdb code). Why should we not show the results of these per defaults?
Having this enabled by default would also match what we do with `prettyPrinters` now (the adjustment in the PR would be minimal, switching the default value in package.json and adjusting the changelog to say ", enabled by default").

... @JacquesLucke and if you update it again, can you please include the single type fix open from https://github.com/WebFreak001/code-debug/pull/434/files#diff-15ecb3c20d4d1bb96e24827f055831bc6dad0a83b404fb8fdc32da1e0d7a831f (yopu know, just because you changed that file as well) and we don't need an extra commit for that change)?

GitMensch:
per GDB docs the frame-filters and related commands are in since GDB 7.7 and this is our minimal supported version (I think we should add the GDB 7.7+ / LMDB 3.7+ note to the main README...) so that should be fine.
The Changelog entry will allow users to know about it and the new setting allows to provide the old behavior.

JacquesLucke:
@GitMensch I did what you suggest. However, in my tests the enable-by-default part does not really work, same for pretty-printing btw. That's because `args.frameFilters` and `args.valuesFormatting` are undefined if not explicitly provided. It does not seem to use the default set in `package.json`.

GitMensch:
thanks for the update, please squash the commit into one (if I do that your commit is lost - maybe I do that wrong)

JacquesLucke:
I squashed all my commits into one. I thought Github would let you do this easily as well.

GitMensch:
you're right, I'll try this merge-squash next

Issue #429, opened 2024-06-06
Popup error on hovering.


memchr:
Apparently, hovering over comments causes the extension to try to evaluate the expression, resulting in many pop-up errors

This is similar to #46 but a bit more disruptive.

---

For example, when I hover on word `cursor` from comment `// don't warp cursor outside the window`, it was evaluated in gdb, resulting in the popup stating `no symbol "cursor" in current context`. I tried to suppress this error but couldn't find the appropriate options.

![image](https://github.com/WebFreak001/code-debug/assets/118117622/6361ab1c-1e30-474a-a133-97d211f76118)

---

using gdb 14.2

oltolm:
I fixed it in this PR https://github.com/WebFreak001/code-debug/pull/438.

PR #438, opened 2024-08-24
prevent error in evaluateRequest when a variable is not found

oltolm:
This prevents the error message when a variable can not be found. Fixes https://github.com/WebFreak001/code-debug/issues/429.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Versioning].
   ([@GitMensch])
 - New `frameFilters` option for GDB that allows using custom frame filters,
   enabled by default ([@JacquesLucke])
+- Suppress error for hover as the user may just play with the mouse ([@oltolm]).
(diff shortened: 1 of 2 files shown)


GitMensch:
This should definitely be disabled for 'hover' but the 'watch' context shows the error not in a message box but in the variable "output", no?

oltolm:
I changed it to only send an error in the `watch` context.

GitMensch (changes requested):
That's a good direction. 

Please swap the check testing for hover as this makes it more clear that we don't want an error for this case, an additional comment like "suppress error for hover as the user may just play with the mouse" would be good, too.

Add a Changelog entry and that's good for merge.

oltolm:
Done. Please release a new version. I fixed a severe bug 6 months ago and it is still not in a released version.

PR #439, opened 2024-08-26
1. fix missing output of variable type for structure

HenryRiley0:
before:

image
image

after:
image
image


GitMensch:
please reword to something like "fix missing output of variable type for structs" (or what this actually fixes), can be merged afterwards

Thank you for your contributions!

PR #440, opened 2024-08-31
Add Aliceserver as a debugger target.

dd86k:
This PR implements Aliceserver as a usable debugger target.

[Aliceserver](https://github.com/dd86k/aliceserver), being a work in progress, is a new debugger that implements both the DAP and MI protocols using [Alicedbg](https://github.com/dd86k/alicedbg) at its core.

It currently only supports the single-session mode, and only supports attach and launch requests.

Requests to implement:
- [ ] Thread info

By knowing what requests an established client make, I am able to extend Alicedbg, giving me a better direction for future features.

Feel free to perform a rebase.

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -33,7 +33,7 @@ class LogMessage {
 	protected logReplaceTest = /{([^}]*)}/g;
 	public logMsgBrkList: Breakpoint[] = [];
 
-	logMsgOutput(record){
+	logMsgOutput(record: { isStream?: boolean; type: any; asyncClass?: string; output?: [string, any][]; content: any; }){
 		if ((record.type === 'console')) {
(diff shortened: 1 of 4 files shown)


src/backend/mi2/mi2.ts
GitMensch, line 1: Can you please send a PR for that one to master?

I haven't checked the current related PR, but maybe there are other pieces outside of alice which you could include in the PR.

Thanks!

dd86k: I can make a PR for these changes. However, if I recall correctly, I did this commit following some tool's recommendation, forgot which, and I haven't checked if these changes cause additional issues. These alone do not depend on anything, thankfully.

Aliceserver & Alicedbg are still (super) WIP, so this PR series will be up as a draft for a while... Sorry.

PR #444, opened 2024-09-18
Add `registerLimit` config option to specify the registers to list in Registers view

### LINKED ISSUE #421: Ignore unfetchable registers
- [x] If you are using gdb
  - [x] `gdb --version` >= 7.7.1
  - [x] it works on the command line with `gdb`
  - [x] `cwd` and `target` are properly set

Sometimes a register is listed in GDB but its value is actually not fetchable even using command line gdb (I guess it's a bug of Qemu's GDBStub) . I think in this case, that specific register, rather than the whole register list, should be ignored. 

In my case it's the "pmpcfg1" register that caused the issue.

The error message says: Could not expand variable: Could not fetch register "pmpcfg1"; remote failure reply 'E14' (from data-list-register-values N)

image

chenzhiy2001:
In cases like #421, fetching unfetchable registers can corrupt the whole Registers view. 

So I added `registerLimit` config option to let users specify what registers they want the debugger to automatically fetch, thus avoiding the issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -299,6 +299,11 @@
 										"description": "Content will be executed on the SSH host before the debugger call."
 									}
 								}
+							},
+							"registerLimit": {
+								"type": "string",
+								"description": "List of numbers specifying the registers to display. An empty string indicates that the contents of all the registers must be returned.",
+								"default": ""
 							}
 						}
 					},
@@ -468,6 +473,11 @@
 										"description": "Content will be executed on the SSH host before the debugger call."
 									}
 								}
+							},
+							"registerLimit": {
+								"type": "string",
+								"description": "List of numbers specifying the registers to display. An empty string indicates that the contents of all the registers must be returned.",
+								"default": ""
 							}
 						}
 					}
@@ -766,6 +776,11 @@
 										"description": "Content will be executed on the SSH host before the debugger call."
 									}
 								}
+							},
+							"registerLimit": {
+								"type": "string",
+								"description": "List of numbers specifying the registers to display. An empty string indicates that the contents of all the registers must be returned.",
+								"default": ""
 							}
 						}
 					},
@@ -846,6 +861,11 @@
 								],
 								"description": "Whether debugger should stop at application entry point",
 								"default": false
+							},
+							"registerLimit": {
+								"type": "string",
+								"description": "List of numbers specifying the registers to display. An empty string indicates that the contents of all the registers must be returned.",
+								"default": ""
 							}
 						}
 					}
@@ -992,6 +1012,11 @@
 								"type": "array",
 								"description": "mago commands to run when starting to debug",
 								"default": []
+							},
+							"registerLimit": {
+								"type": "string",
+								"description": "List of numbers specifying the registers to display. An empty string indicates that the contents of all the registers must be returned.",
+								"default": ""
 							}
 						}
 					},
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -1143,4 +1173,4 @@
 		"publisherDisplayName": "WebFreak",
 		"publisherId": "e054f80a-50f9-4d80-85c8-6ff87eef6c35"
 	}
-}
+}
```
GitMensch, line 1176: please revert that missing EOL

src/backend/mi2/mi2.ts
```diff
@@ -883,7 +883,7 @@ export class MI2 extends EventEmitter implements IBackend {
 	async getRegisterValues(): Promise<RegisterValue[]> {
 		if (trace)
 			this.log("stderr", "getRegisterValues");
-		const result = await this.sendCommand("data-list-register-values N");
+		const result = await this.sendCommand("data-list-register-values N"+ " " +this.registerLimit);
```
GitMensch, line 886: `const result = await this.sendCommand("data-list-register-values N " + this.registerLimit);`
but maybe we should use
`const result = await this.sendCommand("data-list-register-values --skip-unavailable N " + this.registerLimit);`?

GitMensch (changes requested):
Thank you for this nice PR!

Apart from the comments below:
* please add an entry to the Changelog file
* it is likely reasonable to note this option in the README, as well as pointing out how to get the numbers "(starting from zero)", which I guess is either `info registers` or `interpreter mi2 "-data-list-register-values --skip-unavailable N"`.

GitMensch (approved):

GitMensch:
spell check linter failed on "unfetchable", I've asked chatgpt4 for an adjusted text and it changed some other pieces as well, coming up with (after some hinting)

```
There is a Registers view in the VARIABLES view. As we fetch all registers at once, there can be cases where a register that cannot be fetched causes the entire register request to fail, corrupting the entire Registers output. If this happens, you might need to set the `registerLimit` option to specify which registers you want the debugger to fetch automatically.  
For example, to display only registers `rax` and `rip` in an x64 debug session, send the command `-data-list-register-names` in the debug console of an active x64 debug session. You will then receive a response containing an array starting with `["rax","rbx" ...]`. In this array, the index of `rax` is 0 and `rip` is 16, so set the option as `"registerLimit": "0 16"`. If you find the response text hard to navigate, you can paste it into a browser's developer tools console and press enter to get an expandable response object with array elements' indices explicitly displayed.
```

chenzhiy2001 (commented):
- Reverted the deletion of EOL in package.json
- Appended `--skip-unavaliable` after `-data-list-register-names`
- Added an entry in CHANGELOG.md
- Added an introduction of how to set `registerLimit` option in README.md

chenzhiy2001:
The introduction paragraph in README.md is being swapped by the new version now.

GitMensch:
I've run the prettier stuff locally, that's the diff you need to apply:;

```diff

chenzhiy2001:
Oops, I forgot to run `npm run prettier-write-docs`. Could you please reopen this pull request so that the correct modification of CHANGELOG.md show up here?

GitMensch:
I have no clue why the close happened...

chenzhiy2001:
> I have no clue why the close happened...

Sometimes Github gets really confusing. I also felt that.

Issue #75, opened 2016-09-13
supportsRunInTerminalRequest


WebFreak001:
There is a new feature to run the debugger in the vscode integrated console in vscode 1.5

a1994846931931:
Hey, this is a great ideal to support the VSCode integrated terminal, which is of vital importance to debug programs with standard input. I think I need this feature badly. 

But do we have any plans for this?

PR #447, opened 2024-09-25
add integrated terminal support

HenryRiley0:
wechat_2024-09-24_170434_850

Set the default input and output in the VSCode terminal, SSH is not supported for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ Versioning].
 - solve the problem of failed parsing of containers ([@henryriley0])
 - Fixes #421 - Added `registerLimit` option to specify the registers to
   display - PR #444 ([@chenzhiy2001])
+- add integrated terminal support ([@henryriley0])
 
 ## [0.27.0] - 2024-02-07
 
--- a/src/backend/linux/console.ts
+++ b/src/backend/linux/console.ts
@@ -4,7 +4,7 @@ import * as fs from "fs";
 export function spawnTerminalEmulator(preferedEmulator: string): Thenable<string> {
 	return new Promise((resolve, reject) => {
 		const ttyFileOutput = "/tmp/vscode-gdb-tty-0" + Math.floor(Math.random() * 100000000).toString(36);
-		ChildProcess.spawn(preferedEmulator || "x-terminal-emulator", ["-e", "sh -c \"tty > " + ttyFileOutput + " && sleep 4294967294\""]);
+		ChildProcess.spawn("x-terminal-emulator", ["-e", "sh -c \"tty > " + ttyFileOutput + " && sleep 4294967294\""]);
 		let it = 0;
 		const interval = setInterval(() => {
 			if (fs.existsSync(ttyFileOutput)) {
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -6,6 +6,7 @@ import * as linuxTerm from '../linux/console';
 import * as net from "net";
 import * as fs from "fs";
 import * as path from "path";
+import { DebugProtocol } from 'vscode-debugprotocol';
 import { Client, ClientChannel, ExecOptions } from "ssh2";
 
 export function escape(str: string) {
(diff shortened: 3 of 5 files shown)


src/mibase.ts
```diff
@@ -776,7 +776,80 @@ export class MI2DebugSession extends DebugSession {
...
+					`#!/usr/bin/env sh
+					# reset terminal to clear any previous states
+					reset
+					echo "The input and output of program will be here."
+					echo "Warning of set controlling terminal fail can be ignored."
+					tty > ${ttyTmpDir}/ttynameTmp
+					mv ${ttyTmpDir}/ttynameTmp ${ttyTmpDir}/ttyname
+					# wait for debug to finish
+					# prefer using tail to detect PID exit, but that requires GNU tail
+					tail -f --pid=${process.pid} /dev/null 2>/dev/null || while kill -s 0 ${process.pid} 2>/dev/null; do sleep 1s; done
```
GitMensch, line 807: Does this mean you consider the PR unfinished / a working draft?

src/mibase.ts
```diff
@@ -776,7 +776,80 @@ export class MI2DebugSession extends DebugSession {
...
+			try {
+				fs.writeFileSync(
+					`${ttyTmpDir}/get-tty`,
+					`#!/usr/bin/env sh
+					# reset terminal to clear any previous states
+					reset
+					echo "The input and output of program will be here."
+					echo "Warning of set controlling terminal fail can be ignored."
+					tty > ${ttyTmpDir}/ttynameTmp
+					mv ${ttyTmpDir}/ttynameTmp ${ttyTmpDir}/ttyname
```
GitMensch, line 804: Just fo0r understanding: Is the first command slow and the `mv` is used to ensure it is finished before it is available under the name?

src/mibase.ts
```diff
@@ -776,7 +776,80 @@ export class MI2DebugSession extends DebugSession {
...
+		};
+		const ttyTmpDir = await mkdtempAsync(os.tmpdir(), 'debug');
+
+		(async () => {
+			try {
+				fs.writeFileSync(
+					`${ttyTmpDir}/get-tty`,
+					`#!/usr/bin/env sh
+					# reset terminal to clear any previous states
+					reset
```
GitMensch, line 800: This uses the shell's init commands, no?

Should this be suppressed? Should there be an option to add extra commands?

src/backend/mi2/mi2.ts
```diff
@@ -116,19 +117,20 @@ export class MI2 extends EventEmitter implements IBackend {
...
-				if (separateConsole !== undefined) {
+				if (separateConsole == "external") {
 					linuxTerm.spawnTerminalEmulator(separateConsole).then(tty => {
 						promises.push(this.sendCommand("inferior-tty-set " + tty));
 						promises.push(...autorun.map(value => { return this.sendUserInput(value); }));
 						Promise.all(promises).then(() => {
-							this.emit("debug-ready");
+							if(this.application.includes('gdb')){
+								this.emit("debug-ready");
+							}
```
GitMensch, line 127: So no debug-ready for non GDB scenarios and none for win-32 any more?!?

src/backend/linux/console.ts
```diff
@@ -4,7 +4,7 @@ import * as fs from "fs";
 export function spawnTerminalEmulator(preferedEmulator: string): Thenable<string> {
 	return new Promise((resolve, reject) => {
 		const ttyFileOutput = "/tmp/vscode-gdb-tty-0" + Math.floor(Math.random() * 100000000).toString(36);
-		ChildProcess.spawn(preferedEmulator || "x-terminal-emulator", ["-e", "sh -c \"tty > " + ttyFileOutput + " && sleep 4294967294\""]);
```
GitMensch, line 7: Shouldn't we still provide the option to use a preferedEmulator?

src/mibase.ts
```diff
@@ -776,7 +776,80 @@ export class MI2DebugSession extends DebugSession {
 			this.sourceFileMap = new SourceFileMap(configMap, fallbackGDB);
 		}
 	}
+	public async createIntegratedTerminalLinux(args:DebugProtocol.RunInTerminalRequestArguments) {
+		const mkdirAsync = fs.promises.mkdir;
+		const mkdtempAsync = async (tempDir: string, prefix: string): Promise<string> => {
+			const name = `${prefix}-${Date.now()}-${Math.floor(Math.random() * 1e9)}`;
```
GitMensch, line 782: What about using the same approach as above for the tempName?
`Math.floor(Math.random() * 100000000).toString(36)`

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ Versioning].
 - solve the problem of failed parsing of containers ([@henryriley0])
 - Fixes #421 - Added `registerLimit` option to specify the registers to
   display - PR #444 ([@chenzhiy2001])
+- add integrated terminal support ([@henryriley0])
```
GitMensch, line 27: please reference the issue #75 and/or PR #447 here as well

GitMensch (commented):
Please comment on the current state . I'm not sure if this is expected to work on win32 and if it breaks any scenarios (for example is ssh broken or just not work with the internal terminal)?

src/backend/mi2/mi2.ts
HenryRiley0: For now, that's the situation. In the future, we can try to add more, but it might involve more work.

src/backend/linux/console.ts
HenryRiley0: ok

src/mibase.ts
HenryRiley0: the first command (tty > ${ttyTmpDir}/ttynameTmp) is not inherently slow. The use of mv here is not primarily about addressing slowness but rather about ensuring atomicity and consistency

PR #448, opened 2024-09-25
Resolve the issue of not being able to set the GDB path on Windows.

HenryRiley0:
wechat_2024-09-25_162649_083

### DIFF
--- a/src/mibase.ts
+++ b/src/mibase.ts
@@ -96,9 +96,15 @@ export class MI2DebugSession extends DebugSession {
 	// verifies that the specified command can be executed
 	protected checkCommand(debuggerName: string): boolean {
 		try {
-			const command = process.platform === 'win32' ? 'where' : 'command -v';
-			execSync(`${command} ${debuggerName}`, { stdio: 'ignore' });
-			return true;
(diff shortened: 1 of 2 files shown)


src/mibase.ts
```diff
@@ -96,10 +96,16 @@ export class MI2DebugSession extends DebugSession {
...
+			if (process.platform === 'win32' && debuggerName.includes("\\")) {
+				const command = 'dir';
+				execSync(`${command} ${debuggerName}`, { stdio: 'ignore' });
+				return true;
+			}
+			else {
+				const command = process.platform === 'win32' ? 'where' : 'command -v';
+				execSync(`${command} ${debuggerName}`, { stdio: 'ignore' });
+				return true;
+			}		} catch (error) {
```
GitMensch, line 108: I think that `catch` line misses it linebreak, but other than that good to go; suggestion only: "not being able to set the GDB path" -> "not being able to set the GDB binary with a path"

HenryRiley0: Thanks for the correction, it has been revised. However, there is a dead link check failure in the lint_docs self - inspection.

GitMensch:
Thanks for coming back to this.

Issue #413, opened 2024-03-02
Variables Evaluate is NOT Working for Struct Pointer of Function Variable


xiaoyun94:
If submitting a bug please make sure

- [Y] If you are using gdb
  - [Y] `gdb --version` >= 7.7.1
  - [Y] it works on the command line with `gdb`
  - [Y] `cwd` and `target` are properly set
- [ ]  If you are using lldb
  - [ ] `lldb --version` >= 3.7.1
  - [ ] it works on the command line with `lldb-mi`
  - [ ] `cwd` and `target` are properly set

Screenshots are helpful but not required
All extension versions matter.
My Gdbs with verison 10.1/9.x have same problem 
![image](https://github.com/WebFreak001/code-debug/assets/8382233/55840671-fcd5-4a7e-bc1f-58d1278ed262)

xiaoyun94:
This PR
https://github.com/WebFreak001/code-debug/pull/414/files
Deal with this problem, but maybe not the best solution

GitMensch:
Thank you for both raising this and providing a PR for the fix. LGTM.
Can you please rebase the PR and add a Changelog entry, as well as providing an "after" picture?

PR #452, opened 2024-12-02
1.add qt nested structure unit test

HenryRiley0:
add qt unit test


GitMensch:
That only "partial QT" related, this is mostly about "complex structure parsing" - can you rename the test / changelog entry before I pull that in?

GitMensch:
and... maybe you can add #413 to that test as well / and/or finish its PR separately by adding a testcase/changelog entry?

PR #456, opened 2025-03-30
Support unix-based X11 server connections as well as TCP-based ones

karcherm:
Typically, on unix systems, local X11 is connected through the socket `/tmp/.X11-unix/X0`, as indicated by the missing host name part in `DISPLAY=:0` or `DISPLAY=:0.0`. WSLg (the X11 implementation provided with WSL in Microsoft Windows) does not listen on the TCP socket localhost:6000, but on the Unix socket only.

So this pull request enables SSH X11 forwarding to connect to the unix domain socket if the X server host name is set to the empty string.

### DIFF
--- a/package.json
+++ b/package.json
@@ -281,12 +281,12 @@
 											"number",
 											"string"
 										],
-										"description": "Port to redirect X11 data to (by default port = display + 6000)",
+										"description": "TCP port to redirect X11 data to (by default port = display + 6000); raw display number for unix domain connection",
 										"default": 6000
 									},
 									"x11host": {
 										"type": "string",
-										"description": "Hostname/ip to redirect X11 data to",
+										"description": "Hostname/ip to redirect X11 data to; empty string for a unix domain connection",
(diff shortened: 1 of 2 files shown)


WebFreak001:
you should add some mention about this in the user facing config documentation: https://github.com/WebFreak001/code-debug/blob/eb9f5e4dd7dc64f6e6e8a40115d47d30e522525c/package.json

WebFreak001 (approved):

GitMensch:
Oh, is that possibly a step for the next release? Note: the last one is over a year old and we have a bunch of bugfixes and features, see Changelog (this one still needs to be added) from the diff https://github.com/WebFreak001/code-debug/compare/v0.27.0...master (note: #448 is mostly ready to get in as well).

So @WebFreak001: any chance for a release (which is still done and then published to the stores by yourself), soon?

WebFreak001:
the store publishing should be part of GH actions, no?

PR #457, opened 2025-05-21
Refactor equality checks to use strict equality (===/!==)

quanzhuo:
always use ===/!== instead of ==/!=

### DIFF
--- a/src/backend/gdb_expansion.ts
+++ b/src/backend/gdb_expansion.ts
@@ -15,9 +15,9 @@ const pointerCombineChar = ".";
 export function isExpandable(value: string): number {
 	let match;
 	value = value.trim();
-	if (value.length == 0) return 0;
+	if (value.length === 0) return 0;
 	else if (value.startsWith("{...}")) return 2; // lldb string/array
-	else if (value[0] == '{') return 1; // object
+	else if (value[0] === '{') return 1; // object
 	else if (value.startsWith("true")) return 0;
 	else if (value.startsWith("false")) return 0;
 	else if (match = nullpointerRegex.exec(value)) return 0;
@@ -33,7 +33,7 @@ export function isExpandable(value: string): number {
 export function expandValue(variableCreate: (arg: VariableObject | string, options?: any) => any, value: string, root: string = "", extra: any = undefined): any {
 	const parseCString = () => {
 		value = value.trim();
-		if (value[0] != '"' && value[0] != '\'')
+		if (value[0] !== '"' && value[0] !== '\'')
 			return "";
 		let stringEnd = 1;
 		let inString = true;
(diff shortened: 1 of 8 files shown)


src/backend/gdb_expansion.ts
```diff
@@ -66,7 +66,7 @@ export function expandValue(variableCreate: (arg: VariableObject | string, optio
 		stack.push(variable);
 		stack.forEach(name => {
 			prefix = "";
-			if (name != "") {
+			if (name !== "") {
```
WebFreak001, line 69: possible change in behavior here, would need to investigate this area more - can you leave this out this PR for now?

src/backend/gdb_expansion.ts
```diff
@@ -225,21 +225,21 @@ export function expandValue(variableCreate: (arg: VariableObject | string, optio
...
 	createValue = (name: string, val: any) => {
 		let ref = 0;
-		if (typeof val == "object") {
+		if (typeof val === "object") {
 			ref = variableCreate(val);
 			val = "Object";
-		} else if (typeof val == "string" && val.startsWith("*0x")) {
-			if (extra && MINode.valueOf(extra, "arg") == "1") {
+		} else if (typeof val === "string" && val.startsWith("*0x")) {
+			if (extra && MINode.valueOf(extra, "arg") === "1") {
```
WebFreak001, line 232: same here

src/backend/mi2/mi2.ts
```diff
@@ -461,7 +461,7 @@ export class MI2 extends EventEmitter implements IBackend {
 											case "exited-normally":
 												this.emit("exited-normally", parsed);
 												break;
-											case "exited": // exit with error code != 0
+											case "exited": // exit with error code !== 0
```
WebFreak001, line 464: this didn't change, the comment shouldn't change

src/backend/mi2/mi2.ts
```diff
@@ -478,18 +478,18 @@ export class MI2 extends EventEmitter implements IBackend {
...
+								} else if (record.asyncClass === "thread-exited") {
 									this.emit("thread-exited", parsed);
 								}
 							}
 						}
 					});
 					handled = true;
 				}
-				if (parsed.token == undefined && parsed.resultRecords == undefined && parsed.outOfBandRecord.length == 0)
+				if (parsed.token === undefined && parsed.resultRecords === undefined && parsed.outOfBandRecord.length === 0)
```
WebFreak001, line 492: need to also check the types more here to see if this could not happen otherwise, so better to leave this out for now

WebFreak001 (changes requested):
rest looks good

WebFreak001 (approved):

PR #458, opened 2025-05-27
Cache register names

quanzhuo:
Cache register names

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -852,15 +852,13 @@ export class MI2 extends EventEmitter implements IBackend {
 		if (trace)
 			this.log("stderr", "getRegisters");
 
-		// Getting register names and values are separate GDB commands.
-		// We first retrieve the register names and then the values.
-		// The register names should never change, so we could cache and reuse them,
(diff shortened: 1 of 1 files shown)


src/backend/mi2/mi2.ts
```diff
@@ -852,15 +852,13 @@ export class MI2 extends EventEmitter implements IBackend {
 		if (trace)
 			this.log("stderr", "getRegisters");
 
-		// Getting register names and values are separate GDB commands.
-		// We first retrieve the register names and then the values.
-		// The register names should never change, so we could cache and reuse them,
-		// but for now we just retrieve them every time to keep it simple.
-		const names = await this.getRegisterNames();
+		if (!this.regNames) {
+			this.regNames = await this.getRegisterNames();
```
GitMensch, line 856: the idea is good, but we need to store the regnames per inferior (you may switch between multiple ones via UI), not globally, as the inferiors can have different registers (example: server in 64 bit and client in 32 bit (or vice versa), attached to both)

quanzhuo:
Close this PR first. When I'm ready, I'll submit a new one.
Thanks all.

PR #461, opened 2025-11-21
Add multi-process support

JulianvDoorn:
I have rebased https://github.com/WebFreak001/code-debug/pull/133 to the new master since it hasn't been updated. Then I fixed some things to make the extension more respectable to pausing and continuing individual processes. I think only gdb is supported.

Please let me know if the problems that existed in the past are still present. I did some manual fuzzy testing on a C++ file.


JulianvDoorn:
There is a simple C++ program that I used to test. In fact I asked ChatGPT to directly translate https://github.com/WebFreak001/code-debug/pull/133#issuecomment-367473724 to C++ as the D code seemed less stable to me when I started testing. Also I added inferior processes that immediately exit, since this is a particular use case I'm interested in.

```cpp
#include <iostream>
#include <thread>
#include <chrono>
#include <unistd.h>     // fork, sleep
#include <map>
#include <string>
#include <vector>

struct T {
    uint8_t d;
    char e;
    T* child;
};

struct S {
    int a;
    long b;
    std::string c;
    T* f;
};

void fork_and_exit_other();

void secondThread();

int main(int argc, char* argv[]) {
    // Print environment variables
    extern char **environ;
    for (char **env = environ; *env != nullptr; env++) {
        std::string entry = *env;
        auto pos = entry.find('=');
        if (pos != std::string::npos) {
            std::string key = entry.substr(0, pos);
            std::string value = entry.substr(pos + 1);
            std::cout << key << " = " << value << "\n";
        }
    }
    std::cout.flush();

    // Print 0, 1, 2
    for (int i = 0; i < 3; i++) {
        std::cout << i << "\n";
        std::cout.flush();
    }

    // Handle args
    std::vector<std::string> args;
    for (int i = 0; i < argc; i++)
        args.emplace_back(argv[i]);

    if (args.size() > 1 && args[1] == "fork") {
        pid_t origin = getpid();
        pid_t i = fork();

        if (i != 0) {
            secondThread();
        }

        fork_and_exit_other();
        

    } else if (args.size() == 1 || args[1] != "single") {
        std::thread th(secondThread);
        th.detach();
    }

    // Write without newline
    std::cout << "No newline!";
    std::cout.flush();

    int count = 5;
    for (; count < 20; count++) {
        S s;
        s.f = new T();
        s.f->child = new T();
        s.f->child->d = 4;
        s.a = count;
    }

    // Print arguments
    std::cout << "Got Arguments: ";
    for (auto &a : args)
        std::cout << a << " ";
    std::cout << "\n";

    // Main infinite loop
    for (thread_local int i = 0; ; i++) {
        std::cout << "Multithreaded Main thread " << i << "\n";
        std::cout.flush();
        std::this_thread::sleep_for(std::chrono::seconds(1));
    }

    return 0;
}

void fork_and_exit_other() {
    pid_t pid = fork();

    if (pid == 0) {
        exit(1);
    }
}

void secondThread() {
    for (thread_local int i = 0; ; i++) {
        std::cout << "Multithreaded " << i << "\n";
        std::cout.flush();
        std::this_thread::sleep_for(std::chrono::seconds(1));
    }
}
```

JulianvDoorn:
There are still some problems in the combination of threads and processes because the threadId is common among threads and processes. I'm looking for a solution.

JulianvDoorn:
Added initial support for child processes viewer. Worked on auto-discovery of processes and subsequently for each a new 'inferior debug session' is started, necessary for VSCode to create the window as desired. It uses a trick that for every inferior process a new port is opened which redirects events to the original debugger.

image

src/mibase.ts
```diff
@@ -136,50 +146,189 @@ export class MI2DebugSession extends DebugSession {
...
+			console.error('>> accepted connection from client');
+			socket.on('end', () => {
+				console.error('>> client connection closed\n');
+			});
+			const session = new MI2InferiorSession(superiorServer);
+			session.setRunAsServer(true);
+			session.start(socket, socket);
+
+			this.mi2Inferiors.push(session);
+		}).listen(port);
```
WebFreak001: I think you should remove the `port` argument so the OS picks any open port instead.

src/backend/mi2/mi2.ts
```diff
@@ -462,7 +472,7 @@ export class MI2 extends EventEmitter implements IBackend {
 												this.emit("exited-normally", parsed);
 												break;
 											case "exited": // exit with error code != 0
-												this.log("stderr", "Program exited with code " + parsed.record("exit-code"));
+												this.log("stderr", "Inferior exited with code " + parsed.record("exit-code"));
```
WebFreak001: I think this might confuse people who are not that deep into debugger terms. Perhaps it should only log inferior when multiProcess is set

src/backend/mi2/mi2inferior.ts
```diff
@@ -0,0 +1,101 @@
...
+
+	examineMemory(from: number, to: number): Thenable<any> {
+		return this.superior.examineMemory(from, to);
+	}
+	
+	private superior: MI2;
+
+	contructor(superior: MI2) {
+		this.superior = superior
+	}
```
WebFreak001: can you move constructor and field up to the start of the class? I think they are important for understanding the class and are better read before the implementations using the field

src/mibase.ts
```diff
@@ -136,50 +146,189 @@ export class MI2DebugSession extends DebugSession {
 	}
 
 	protected handleBreakpoint(info: MINode) {
+		let threadId = info.record("thread-id");
+
+		this.miDebugger.log("stdout", `Got theadid ${threadId}`);
```
WebFreak001: typo
```suggestion
		this.miDebugger.log("stdout", `Got thread ID ${threadId}`);
```

src/mibase.ts
```diff
@@ -136,50 +146,189 @@ export class MI2DebugSession extends DebugSession {
...
+				}
+			}, 1000, () => {})
+		}
+	}
+
+	protected threadGroupExitedEvent(info: MINode) {
+		let pid = this.threadGroupPids.get(info.record("id"));
+
+		if (pid == this.sessionPid) {
+			this.miDebugger.log("stdout", "this.sesionPid = undefind");
```
WebFreak001: this log is not useful for the user

src/mibase.ts
```diff
@@ -136,50 +146,189 @@ export class MI2DebugSession extends DebugSession {
...
+			const session = new MI2InferiorSession(superiorServer);
+			session.setRunAsServer(true);
+			session.start(socket, socket);
+
+			this.mi2Inferiors.push(session);
+		}).listen(port);
+
+		this.inferiorServers.push(server);
+		
+		return server;
```
WebFreak001, line 261: This return should wait until the port was opened, e.g. return a Promise and resolve it `once` the [listening](https://nodejs.org/api/net.html#event-listening) event

src/mibase.ts
```diff
@@ -136,50 +146,189 @@ export class MI2DebugSession extends DebugSession {
...
+
+		this.miDebugger.log("stdout", `Handling breakpoint for threadPid == this.sessionPid (${threadPid} == ${this.sessionPid})`)
+
+		if (threadPid == this.sessionPid) {
+			this.sendEvent(event);
+		} else {
+			this.mi2Inferiors.forEach(inferior => {
+				if (threadPid == inferior.sessionPid) inferior.sendEvent(event)
+			});
+		}
```
WebFreak001, line 164: this construct is quite a lot of times, it should probably be moved into a utility function.

Also does this even respect the multiProcess configuration flag?

src/mibase.ts
```diff
@@ -136,50 +146,189 @@ export class MI2DebugSession extends DebugSession {
...
 		(event as DebugProtocol.StoppedEvent).body.allThreadsStopped = info ? info.record("stopped-threads") === "all" : true;
-		this.sendEvent(event);
+
+		if (threadPid == this.sessionPid) {
+			this.sendEvent(event);
+		} else {
+			this.mi2Inferiors.forEach(inferior => {
+				if (threadPid == inferior.sessionPid) inferior.sendEvent(event)
+			});
+		}
```
WebFreak001, line 179: ditto

src/mibase.ts
```diff
@@ -136,50 +146,189 @@ export class MI2DebugSession extends DebugSession {
...
 		(event as DebugProtocol.StoppedEvent).body.allThreadsStopped = info.record("stopped-threads") === "all";
-		this.sendEvent(event);
+
+		if (threadPid == this.sessionPid) {
+			this.sendEvent(event);
+		} else {
+			this.mi2Inferiors.forEach(inferior => {
+				if (threadPid == inferior.sessionPid) inferior.sendEvent(event)
+			});
+		}
```
WebFreak001, line 194: ditto

src/mibase.ts
```diff
@@ -136,50 +146,189 @@ export class MI2DebugSession extends DebugSession {
...
 			(event as DebugProtocol.StoppedEvent).body.allThreadsStopped = info.record("stopped-threads") === "all";
-			this.sendEvent(event);
+
+			if (threadPid == this.sessionPid) {
+				this.sendEvent(event);
+			} else {
+				this.mi2Inferiors.forEach(inferior => {
+					if (threadPid == inferior.sessionPid) inferior.sendEvent(event)
+				});
+			}
```
WebFreak001, line 212: ditto

src/mibase.ts
```diff
@@ -136,50 +146,189 @@ export class MI2DebugSession extends DebugSession {
...
+		let threadPid = this.threadGroupPids.get(info.record("group-id"));
+		this.threadToPid.set(threadId, threadPid);
+
+		if (threadPid == this.sessionPid) {
+			this.sendEvent(new ThreadEvent("started", threadId));
+		} else {
+			this.mi2Inferiors.forEach(inferior => {
+				if (threadPid == inferior.sessionPid) inferior.sendEvent(new ThreadEvent("started", threadId));
+			});
+		}
```
WebFreak001, line 228: ditto

src/mibase.ts
```diff
@@ -136,50 +146,189 @@ export class MI2DebugSession extends DebugSession {
...
+		let threadPid = this.threadGroupPids.get(info.record("group-id"));
+		this.threadToPid.delete(info.record("group-id"));
+
+		if (threadPid == this.sessionPid) {
+			this.sendEvent(new ThreadEvent("exited", threadId));
+		} else {
+			this.mi2Inferiors.forEach(inferior => {
+				if (threadPid == inferior.sessionPid) inferior.sendEvent(new ThreadEvent("exited", threadId));
+			});
+		}
```
WebFreak001, line 243: ditto

src/mibase.ts
```diff
@@ -136,50 +146,189 @@ export class MI2DebugSession extends DebugSession {
...
 
-		if (this.serverPath)
-			fs.unlink(this.serverPath, (err) => {
+		this.threadGroupPids.delete(info.record("id"));
+	}
+
+	protected quitEvent(info?: MINode) {
+		if (this.threadGroupPids.size == 0) {
+			this.quit = true;
+			this.sendEvent(new TerminatedEvent());
```
WebFreak001: is this not sent to the inferior sessions?

WebFreak001 (changes requested):
Thanks for the work! Looks pretty good to me, but I can't guarantee any timeframe when this will be merged since I've been only very sporadically looking at my projects recently.

I have looked at this very briefly, I was not able to review the code for more deeply contained bugs or issues yet - but perhaps it's fine to just move them to live testing with users due to constrained reviewing resources.

Got a few requests already though, also can you resolve the linter/formatting issues that github just added in annotations?

Perhaps @GitMensch is interested in this too?

JulianvDoorn:
Thanks for the feedback. Right now the PR is still a WIP, since the scope increased because I went down the rabbit hole of starting multiple debug sessions. I have a very strong feeling that once every works stably, there is a huge opportunity for refactoring which I think is necessary prior to merging.

JulianvDoorn:
Nevertheless, I have relatively little experience in Node.js and TypeScript so I will welcome feedback on style en conventions. Also I also had to add some features in the VSCode debug adapter to support this PR. Once the changes settle I will open a PR over there as well. So it will take a while before this PR can be merged.

src/mibase.ts
JulianvDoorn: This is sent from the debug adapter to VSCode to indicate that the debugging UI can be closed

src/mibase.ts
JulianvDoorn: I've refactored the way that mibase and miinferior works, allowed for much more reduction in the amount of code

JulianvDoorn:
I think I have gotten to the point that it works relatively stable. Let's see if I can test drive this in the coming days 😄

src/mibase.ts
JulianvDoorn: Removed log entirely

src/backend/backend.ts
```diff
@@ -64,8 +64,8 @@ export interface IBackend {
 	start(runToStart: boolean): Thenable<boolean>;
 	stop(): void;
 	detach(): void;
-	interrupt(): Thenable<boolean>;
-	continue(): Thenable<boolean>;
+	interrupt(threadId?: number): Thenable<boolean>;
+	continue(reverse?: boolean, threadId?: number): Thenable<boolean>;
 	next(): Thenable<boolean>;
```
JulianvDoorn, line 69: This interface does not correspond with the actual implementation

src/mibase.ts
```diff
@@ -135,50 +141,153 @@ export class MI2DebugSession extends DebugSession {
...
 	protected quitEvent() {
 		this.quit = true;
-		this.sendEvent(new TerminatedEvent());
+		this.sendEvent(new ExitedEvent(0));
 
 		if (this.serverPath)
 			fs.unlink(this.serverPath, (err) => {
-				// eslint-disable-next-line no-console
-				console.error("Failed to unlink debug server");
+			// eslint-disable-next-line no-console
```
JulianvDoorn: Missing Indendentation

src/mibase.ts
```diff
@@ -26,44 +29,47 @@ class VariableScope {
...
+		this.shared.miDebugger.on("step-end", this.handleBreak.bind(this));
+		//this.shared.miDebugger.on("step-out-end", this.handleBreak.bind(this));  // was combined into step-end
+		this.shared.miDebugger.on("step-other", this.handleBreak.bind(this));
+		this.shared.miDebugger.on("signal-stop", this.handlePause.bind(this));
+		this.shared.miDebugger.on("thread-created", this.threadCreatedEvent.bind(this));
+		this.shared.miDebugger.on("thread-exited", this.threadExitedEvent.bind(this));
+		this.shared.miDebugger.once("debug-ready", (() => this.sendEvent(new InitializedEvent())));
+		this.shared.miDebugger.on("thread-group-started", this.threadGroupStartedEvent.bind(this));
+		this.shared.miDebugger.on("thread-group-exited", this.threadGroupExitedEvent.bind(this));
+		this.sendEvent(new InitializedEvent());
```
JulianvDoorn: This line seems to break ssh connections. Removing it fixed it but I need to access its impact

PR #464, opened 2026-07-20
Update gdb_expansion.ts

xiaoyun94:
Resolved the issue that the input parameter value of a function cannot be displayed when the structure pointer is used as the input parameter of the function.


xiaoyun94:
Before:
Image

Now: 
image

src/backend/gdb_expansion.ts
```diff
@@ -229,7 +229,7 @@ export function expandValue(variableCreate: (arg: VariableObject | string, optio
 			ref = variableCreate(val);
 			val = "Object";
 		} else if (typeof val === "string" && val.startsWith("*0x")) {
-			if (extra && MINode.valueOf(extra, "arg") == "1") {
+			if (extra && MINode.valueOf(extra, "type") == "char **") {
 				ref = variableCreate(getNamespace("*(" + name), { arg: true });
```
GitMensch, line 233: the previous version checked if it is an arg - and showed it that way + got the content, now that is always done if the type is a `char **`...

Do we need to split that (haven't checked what arg:true does)?

Also: is it possible to add a testcase for that to ensure a follow-up commit does not break it?

